PR #34, opened 2020-07-17
Name conflict resolution with reserved Dart Keywords

### LINKED ISSUE #21: dart keywords as identifiers
What happens when we have dart keywords as identifiers? Keywords: https://dart.dev/guides/language/language-tour#keywords.

We should have warnings, and maybe a workaround/setting.

Here is some related code: https://github.com/dart-lang/discoveryapis_generator/blob/master/lib/src/namer.dart

mannprerak2:
closes #21 

- Added reserved keyword conflict resolution to UniqueNamer
- Added tests (`collision_tests/reserved_word_collision_test.dart`)
- Bug fix: UniqueNamers used by declarations are now reset to initial state before generating code. This ensures multiple calls to Library.generate will produce the same results


lib/src/code_generator/utils.dart
```diff
@@ -2,10 +2,15 @@ import 'dart_keywords.dart';
...
 class UniqueNamer {
   final Set<String> _usedUpNames;
+
+  /// Creates a UniqueNamer with given [usedUpNames] and Dart reserved keywords.
   UniqueNamer(Set<String> usedUpNames)
       : assert(keywords.intersection(usedUpNames).isEmpty),
         _usedUpNames = {...keywords, ...usedUpNames};
 
+  /// Creates a UniqueNamer with given [usedUpNames] only.
+  UniqueNamer.raw(this._usedUpNames);
```
dcharkes, line 12: Make this private?

PR #35, opened 2020-07-21
Added support for parsing macros and anonymous unnamed enums

### LINKED ISSUE #12: Support for constants defined by MACROS
We should investigate whether it's possible to export constants defined by macros.

For example [sqlite3.h](https://www.sqlite.org/android/file?name=sqlite3/src/main/jni/sqlite/sqlite3.h):

```c
#define SQLITE_LIMIT_LENGTH                    0
#define SQLITE_LIMIT_SQL_LENGTH                1
#define SQLITE_LIMIT_COLUMN                    2
#define SQLITE_LIMIT_EXPR_DEPTH                3
#define SQLITE_LIMIT_COMPOUND_SELECT           4
#define SQLITE_LIMIT_VDBE_OP                   5
#define SQLITE_LIMIT_FUNCTION_ARG              6
#define SQLITE_LIMIT_ATTACHED                  7
#define SQLITE_LIMIT_LIKE_PATTERN_LENGTH       8
#define SQLITE_LIMIT_VARIABLE_NUMBER           9
#define SQLITE_LIMIT_TRIGGER_DEPTH            10
#define SQLITE_LIMIT_WORKER_THREADS           11
```

See migration experiment at https://dart-review.googlesource.com/c/sdk/+/152141 (the rest works like a charm).

### LINKED ISSUE #50: Generate bindings for unnamed enums
It is very common to use unnamed enum in C:
```c
enum {
  ENUM_VALUE = 1,
  ENUM_VALUE = 2,
  //...
};
```
please consider generating bindings for them.

mannprerak2:
closes #12, closes #50 
- Support for parsing macros and anonymous unnamed enums.
- These are generated as top-level constants.
- unnamed enums can be disabled by using `unnamed-enums: false`, default is `true`.
- Added and Updated tests.

Macros are saved while parsing headers initially, and are parsed again at the end by creating a file with all headers and creating variables (using auto keyword) which are assigned to macros, this variable's value is then evaluated using libclang.
E.g
```C
#define ABC 10
```
For getting the value of this macro we generate an `hpp` file
```hpp
/// Add all included headers above `#include "location"`
auto _3_ABC_macro_variable = ABC;
```
This file is then parsed again to get type and value of macro.


lib/src/header_parser/sub_parsers/macro_parser.dart
```diff
@@ -0,0 +1,226 @@
...
+/// Base name of generated file.
+String _generatedFileBaseName;
+
+/// Generated macro variable names.
+///
+/// Used to determine if macro should be included in bindings or not.
+Set<String> _macroVarNames = {};
+
+/// Creates a temporary file for parsing macros in current directory.
+File createFileForMacros() {
```
dcharkes, line 174: Should we move this to .dart_tool as well?

lib/src/header_parser/clang_bindings/clang_bindings.dart
```diff
@@ -1706,6 +1837,24 @@ class CXChildVisitResult {
   static const int CXChildVisit_Recurse = 2;
 }
 
+class CXEvalResultKind {
```
dcharkes, line 1840: Maybe make it abstract it's clear that it only contains consts.

dcharkes, line 1840: Generate a comment that this is from an enum, and that the actual enum values are exposed in Dart as integers for functions and struct fields.

lib/src/header_parser/sub_parsers/macro_parser.dart
mannprerak2: This file is temporarily created and then deleted immediately (this happens really fast) so it doesn't matter much.

The issue with changing the location of this file is that all the headers supplied by the user will also need to be updated to use the correct path. 
Currently, the file we generate pastes the headers as it is
```hpp
#include "header1/path/pasted/as/is"
#include "header2/path/pasted/as/is"

auto abc = ABC;
auto efg = EFG;
```

lib/src/header_parser/clang_bindings/clang_bindings.dart
mannprerak2: I am not really sure what you mean.
Do you want to change how enums are generated by our tool?

lib/src/header_parser/sub_parsers/macro_parser.dart
mannprerak2: Well not really an issue, this can be done but it seems unnecessary to me. I can make this change if you want so.

lib/src/header_parser/clang_bindings/clang_bindings.dart
mannprerak2: I am not really sure what you mean.
This enum is part of libclang itself, not our wrapper, so we cannot add a comment to a particular enum.

dcharkes: I meant `abstract class EnumName { ... }`.

lib/src/header_parser/sub_parsers/macro_parser.dart
dcharkes: The current way sounds fine to me.

lib/src/header_parser/clang_bindings/clang_bindings.dart
dcharkes: Ah, yeah of course. I was thinking because this enum value did not have a doc-comment copied from C we could generate explanatory comments. But having both a doc-comment and a generated comment is going to be very confusing for people.

Then just leave it as is.

test/header_parser_tests/macros.h
```diff
@@ -0,0 +1,11 @@
+#define TEST1 1.1
+#define TEST2 10
+#define TEST3 (TEST1 + TEST2)
+#define TEST4 "test"
+
+// This should have been ignored but is parsed as an int of value 1.
```
dcharkes, line 6: I would not expect this to be ignored, I would expect this to be evaluated as a 'comma expression'.

```c++
#define TEST5 4, \
              2, \
              3

int bla3(){
    return TEST5; // returns 3.
}
```

Please check support for comma expressions and support for `\` in expressions separately.

And if this is a bug in llvm/clang, file an issue there, and an issue here pointing to that issue.

test/large_integration_tests/_expected_sqlite_bindings.dart
```diff
@@ -9856,6 +9856,887 @@ class fts5_tokenizer extends ffi.Struct {}
 
 class fts5_api extends ffi.Struct {}
 
+const String SQLITE_VERSION = '3.32.3';
```
dcharkes, line 9859: 🎉

lib/src/strings.dart
```diff
@@ -7,7 +7,7 @@ import 'package:ffigen/src/header_parser/clang_bindings/clang_bindings.dart'
     as clang;
 
 // This version must be updated whenever we update the libclang wrapper.
-const dylibVersion = 'v1';
+const dylibVersion = 'v2';
```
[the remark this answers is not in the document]
mannprerak2: I don't think we will need to change the wrapper anytime soon now.
But I suppose we can find and parse the `pubspec.lock` file to get the exact version resolved and use that.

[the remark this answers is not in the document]
dcharkes: Feel free to file a new issue to address this instead of doing it in this PR.

test/header_parser_tests/macros.h
mannprerak2: Just tested this out,
```c
int a = bla3(); // this returns 3 as expected.
int b = TEST5; // this however doesn't compile and libclang returns the first value in the comma sequence.
```

To make this compile I tried this
```c
#define TEST5 1,2,3

int bla3(){
    return TEST5; // returns 3.
}
int main()
{
    int a = bla3();
    int b;
    b = TEST5; // this compiles now
    printf("%d %d\n", a, b);

    return 0;
}
```
GCC Output: `3 1`

So I think this comma operator is behaves differently incase of functions or variables.

PR #37, opened 2020-07-23
Pub score fix

mannprerak2:


pubspec.yaml
```diff
@@ -3,9 +3,9 @@
 # BSD-style license that can be found in the LICENSE file.
 
 name: ffigen
-version: 0.1.0
+version: 0.1.1
 homepage: https://github.com/dart-lang/ffigen
-description: Experimental generator for FFI bindings.
+description: Experimental generator for FFI bindings, uses LibClang to parse C/C++ header files.
```
dcharkes, line 8: uses -> using

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 0.1.1
+- Fix pub score.
```
dcharkes, line 2: - Address pub score: follow dart File conventions, provide documentation, and pass static analysis.

lib/ffigen.dart
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2020, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// A bindings generator for dart.
+///
+/// See complete usage at - https://pub.dev/packages/ffigen.
+library ffigen;
+
+export 'src/code_generator.dart';
```
dcharkes, line 10: You might want to be particular about what you allow to export from these three files. Once people start using this, you cannot change it anymore.

mannprerak2: Ah, right.
Updated to expose just the things required to build your own `bin/ffigen.dart` script. If the user wants even more flexibility they can always import from `lib/src`.

PR #48, opened 2020-07-24
Update config specification for Comments

### LINKED ISSUE #38: Support for generating comments sqlite3
We talked about supporting non-doc-string comments, we should implement it.

mannprerak2:
Closes #38 
- Updated readme to use `code` tag inside `pre` tag in config options
- Updated comment type in config
- Comments config now has a style and length sub keys - `style: doxygen(default) | any`, `length: brief | full(default)`, and can be disabled by passing `comments: false`.


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 0.1.4
+- Comments config now has a style and length sub keys - `style: doxygen(default) | any`, `length: brief(default) | full`, and can be disabled by passing `comments: false`.
```
dcharkes, line 2: Is there any reason to not make `full` the default? (It is very nice to know all individual parameters when coding against the generated bindings in Dart.)

README.md
```diff
@@ -128,53 +128,57 @@ functions:
...
   <tr>
     <td>comments</td>
     <td>no</td>
     <td>Extract documentation comments for declarations.<br>
-      <i>Options: brief/full/none</i><br>
-      <b>Default: brief</b><br>
-      By default clang only parses documentation comments. To enable
-      ordinary comments (starting with // or /*) add<br>
-      <code>-fparse-all-comments</code> to <code>compiler-opts</code>.
+    You can specify the style and length of the comment from one of these options<br>
```
dcharkes, line 145: The style and length of the comments can be specified with the following options. (remove you)

CHANGELOG.md
mannprerak2: We can, but then the comment is not going to be wrapped.

mannprerak2: Changing it to full then.

dcharkes: I thought we had manual wrapping for doxygen comments?

mannprerak2: The wrapping is only done for brief comments.
For full comments, We just remove the raw comment markups from every line.
This is because the simple wrapping we use can distort code samples.

dcharkes: Ah yeah, of course. That's fine. If the source has proper wrapping, then the generated files should also be okay. (Maybe a couple of characters over the 80 char limit, but that's worth it to preserve to good formatting in my opinion.)

PR #60, opened 2020-07-30
Update headers config.

### LINKED ISSUE #52: Header-filter exclude allow full path
Atm one can define the header-filter excludes using the filename. Though filenames could be ambigious over different folders. Could we use full paths to exclude certain headers?

mannprerak2:
Closes #52 
- Updated headers config to have sub-fields `entry-points` and `include-directives`, removed `header-filter` config.
- Globs are supported by both subkeys.
- Updated tests, examples, readme, changelog.


lib/src/config_provider/config.dart
```diff
@@ -141,21 +136,12 @@ class Config {
         extractor: outputExtractor,
         extractedResult: (dynamic result) => output = result as String,
       ),
-      strings.headers: Specification<List<String>>(
-        description: 'List of C headers to generate bindings of',
+      strings.headers: Specification<Headers>(
+        description: 'Headers.',
```
dcharkes, line 140: Are we exposing this description anywhere? (Like in `--help` or when the config is not valid?)

If yes, then maybe elaborate instead of making the description smaller.
If no, should we expose this documentation somewhere? Or should we remove it?

lib/src/config_provider/config_types.dart
```diff
@@ -5,6 +5,7 @@
 /// Contains all the neccesary classes required by config.
 
 import 'package:meta/meta.dart';
+import 'package:quiver/pattern.dart' as quiver;
```
dcharkes, line 8: 😄

test/native_test/native_test_bindings.dart
```diff
@@ -3,244 +3,180 @@
...
-}
-
-final _dart_Function1IntPtr _Function1IntPtr =
-    _dylib.lookupFunction<_c_Function1IntPtr, _dart_Function1IntPtr>(
-        'Function1IntPtr');
-
-typedef _c_Function1IntPtr = ffi.IntPtr Function(
-  ffi.IntPtr x,
-);
+class NativeLibrary {
```
dcharkes, line 6: I'm confused with the diff of this file, this looks related to a previous PR.

Didn't our tests fail because we did not generate the correct file?

Please ensure we catch this in the future.

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 0.1.6
```
dcharkes, line 1: > Although semantic versioning doesn’t promise any compatibility between versions prior to 1.0.0, the Dart community convention is to treat those versions semantically as well. The interpretation of each number is just shifted down one slot: going from 0.1.2 to 0.2.0 indicates a breaking change, going to 0.1.3 indicates a new feature, and going to 0.1.2+1 indicates a change that doesn’t affect the public API. For simplicity’s sake, avoid using + after the version reaches 1.0.0.

https://dart.dev/tools/pub/versioning#semantic-versions

Let's make all breaking changes increase the minor version number (as long as we're not yet at 1.0).

mannprerak2: Yeah

lib/src/config_provider/config.dart
mannprerak2: Currently, it's not exposed anywhere, but we can add these to --help

mannprerak2: I am removing these descriptions for now.

pubspec.yaml
```diff
@@ -3,7 +3,7 @@
 # BSD-style license that can be found in the LICENSE file.
 
 name: ffigen
-version: 0.1.5
+version: 0.2.0
```
dcharkes, line 6: I think this should be something like `0.2.0-dev` for commits that are not `0.2.0`.

https://semver.org/spec/v2.0.0-rc.1.html#spec-item-10

mannprerak2: Updated.

PR #62, opened 2020-07-31
Simpler include/exclude, regexp based renaming for declarations

### LINKED ISSUE #56: Regexp replacement for renaming declarations
Currently, we support prefix-replacement for declarations, this should be replaced with something a _bit more powerful_ like a regexp replacement solution.

mannprerak2:
Closes #56 
- Removed the use of `names` and `matches` for `include`/`exclude` of declaration.
- Added declaration renaming using `rename` config key.
- Updated tests, examples. Added rename test
- Removed `prefix-replacement` and `prefix` config keys.

Include/Exclude/Renaming for declarations (func, struct, enum, macro) can be done like this
```yaml
function:
  include:
    - 'clang_.*'
  exclude:
    - 'CX_.*'
  rename:
    clang_dispose_string: clangDisposeString
    'clang_(.*)': '$1'
    '_(.*)': '$1'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,8 @@ Add configurations to Pubspec File:
 ffigen:
   output: 'generated_bindings.dart'
   headers:
-    - 'example.h'
+    entry-points:
+      - 'example.h'
 ```
 Output (_generated_bindings.dart_).
 ```dart
@@ -39,7 +40,7 @@ typedef _dart_sum = int Function(int a,int b);
 - Run the tool- `pub run ffigen`.
 
 ## Setup
-`package:ffigen` uses LLVM. Install LLVM in the following way.
+`package:ffigen` uses LLVM. Install LLVM (9+) in the following way.
 
 #### ubuntu/linux
 1. Install libclangdev - `sudo apt-get install libclang-dev`.
--- a/README.md
+++ b/README.md
@@ -107,17 +107,16 @@ headers:
     <td>Filters for declarations.<br><b>Default: all are included</b></td>
     <td><pre lang="yaml"><code>
 functions:
-  include: # Exclude is also available.
-    names: # Matches with exact name.
-      - someFuncName
-      - anotherName
-    matches: # Matches using regexp.
-      - prefix.*
-      - [a-z][a-zA-Z0-9]*
-  prefix: 'cx_' # Prefix added to all functions.
-  prefix-replacement: # Replaces a functions's prefix.
-    'clang_': ''
-    '_': 'C'</code></pre></td>
+  include: # 'exclude' is also available.
+    - [a-z][a-zA-Z0-9]* # Matches using regexp.
+    - prefix.* # '.' matches any character.
+    - someFuncName # Matches with exact name
+    - anotherName # full names have higher priority.
+  rename:
+    'clang_(.*)': '$1' # Regexp groups based replacement.
+    'clang_dispose': 'dispose' # full name matches have higher priority.
+    '_(.*)': '$1' # Removes '_' from beginning of a name.
+    </code></pre></td>
   </tr>
   <tr>
     <td>array-workaround</td>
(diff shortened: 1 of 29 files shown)


README.md
```diff
@@ -107,17 +108,16 @@ headers:
...
-    matches: # Matches using regexp.
-      - prefix.*
-      - [a-z][a-zA-Z0-9]*
-  prefix: 'cx_' # Prefix added to all functions.
-  prefix-replacement: # Replaces a functions's prefix.
-    'clang_': ''
-    '_': 'C'</code></pre></td>
+  include: # 'exclude' is also available.
+    - [a-z][a-zA-Z0-9]* # Matches using regexp.
+    - prefix.* # '.' matches any character.
```
jonasfj, line 113: Wouldn't this also match `blablablaprefix.cheese`? because there is no anchors? From what I see you're using `firstMatch`..

README.md
```diff
@@ -107,17 +108,16 @@ headers:
...
+    - someFuncName # Matches with exact name
+    - anotherName # full names have higher priority.
```
jonasfj, line 115: Are regexes automatically anchored with `^` and `$` or should I do that myself? It might be wise to give an example of this...

Also, how do you tell a regex from a full name? `anotherName` is a valid regex.. I would suggest either doing:
```yaml
include:
  names:
   - fullMethodName
  patterns:
  - ^prefix.*$
```
like before or:
```yaml
include:
  - fullMethodName
  - regex: ^prefix.*$  # Using an inline object with a key 'regex' to make a regular expression
```
or simply only regexes:
```yaml
include:
  - ^fullMethodName$
  - ^prefix.*$
```

Note. if anchors `^$` are not always automatically inserted around the patterns, I would suggest your examples highlights this. Similarly, if they are inserted around the patterns, the documentation should probably highlight this.

README.md
mannprerak2: Are anchors really required?
I am doing a fullMatch [(package:quiver](https://pub.dev/packages/quiver))
- For include/exclude -> https://github.com/mannprerak2/ffigen/blob/declaration_include_exclude_rename/lib/src/config_provider/config_types.dart#L183
 - For renaming, I first see if its a full match, and then use firstMatch (which should return the largest match I suppose)
https://github.com/mannprerak2/ffigen/blob/declaration_include_exclude_rename/lib/src/config_provider/config_types.dart#L126

Also we match with declaration names, which in C can only contain `_, a-z, A-Z, 0-9`

README.md
mannprerak2: > Are regexes automatically anchored

Nope, Do you think this should be necessary? I have not had any trouble using regexp matching till now in any examples or tests, when do you think will this be necessary?

>how do you tell a regex from a full name

We have actually had a pretty long discussion on this, and the final conclusion was to keep no subkeys for include/exclude. We do a simple regexp match to check if a regexp string only consists of `_, a-z, A-Z, 0-9` and if it does, we treat it as a full name, which has performance benefits because we can match in O(1) using a HashSet.

I do not have an in-depth knowledge of RegularExpressions, also we'd like to keep the regexp patterns simple enough, and its possible for us because the C declaration names we match with themselves have a limited pattern set.

README.md
mannprerak2: Here is the function used for full matching (in package quiver)
```dart
bool matchesFull(Pattern pattern, String str) {
  var match = pattern.matchAsPrefix(str);
  return match != null && match.end == str.length;
}
```

README.md
mannprerak2: Just tested it with anchors on the `libclang-example`
```yaml
  structs:
    include:
        - '^CX.*$'
  enums:
    include:
      - '^CXTypeKind$'
```

Using or not using `^` `$` doesn't affect the output

README.md
jonasfj: > Are anchors really required?

No probably not, my bad :D

README.md
jonasfj: > We have actually had a pretty long discussion on this, and the final conclusion was to keep no subkeys for include/exclude. We do a simple regexp match to check if a regexp string only consists of _, a-z, A-Z, 0-9 and if it does, we treat it as a full name, which has performance benefits because we can match in O(1) using a HashSet.

Is performance a concern... if not I would focus on keep the code simple rather than fast :see_no_evil: 

In any case, I see that for the patterns you have this makes sense!

PR #66, opened 2020-08-05
Regexp based Member renaming

### LINKED ISSUE #28: Prefix and replacement for struct/enum members.
We currently have no way to add a prefix/replacement name for a struct/enum member.
That means members starting with `_` will be considered private and thus be unusable in our bindings.

mannprerak2:
Closes #28 
- Added Member renaming using `member-rename` subkey in declarations.
- Renames struct/enum members and function parameter names using regexp/full matching.
- Added/Updated renaming tests.

Renaming can be done like this (example) -
```yaml
enums:
  member-rename:
    '(.*)': # Matches any enum.
      '_(.*)': '$1' # Removes '_' from beginning enum member name.
    'CXTypeKind': # Full names have higher priority.
      'CXType(.*)': '$1' # $1 keeps only the 1st group i.e '(.*)'.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 - Updated header config. Header `entry-points` and `include-directives` are now specified under `headers` key. Glob syntax is allowed.
 - Updated declaration `include`/`exclude` config. These are now specified as a list.
 - Added Regexp based declaration renaming using `rename` subkey.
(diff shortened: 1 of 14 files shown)


mannprerak2:
@jonasfj Could you also merge this?
Commit message-
```
Closes #28 
- Added Member renaming using `member-rename` subkey in declarations.
- Renames struct/enum members and function parameter names using regexp/full matching.
- Added/Updated renaming tests.

```

PR #78, opened 2020-08-12
Update readme - added FAQ, fixed config table

mannprerak2:
- Added FAQ section
- Fixed config table width of columns 2
- Removed `<code>` element from `<pre>` causing indentation error on pub


README.md
```diff
@@ -294,3 +304,58 @@ class ArrayHelper_CXFileUniqueID_data_level0 {
   2. Run `dart build_test_dylib.dart`.
 
 Run tests from the root of the package with `pub run test`.
+
+## FAQ
+<details><summary>Removing underscores or renaming declarations.</summary>
```
dcharkes, line 309: Maybe just use headers instead of details/summary.

dcharkes, line 309: This should be a question, it's a FAQ after all.

### Can ffigen remove underscores or rename declarations?

(same for other headers)

README.md
```diff
@@ -66,20 +68,23 @@ The following configuration options are available-
 <thead>
   <tr>
     <th>Key</th>
-    <th>Explaination</th>
+    <th>
+    <!-- Adding &nbsp to increase column width. -->
+    Explaination&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp
```
dcharkes, line 73: This makes the Explanation be aligned left while the rest is aligned right.
If you want to increase the column with you should use proper html for that, not nbsp characters.

README.md
mannprerak2: I think it might be better to use collapsible details/summary for this because maybe this list can grow a lot over time.

README.md
mannprerak2: > If you want to increase the colum

Updated to use css.
Earlier I tried to do it with CSS but was testing it on vscode's markdown preview where it wasn't working, but it's working on Github so that's fine.

README.md
dcharkes: It's the bottom of the page, so I think it's fine.

Also, if people use ctrl+f to search for something, how does that work if they search for a keyword that is in the collapsed part?

mannprerak2: - removed details/summary

Search doesn't work for collapsed section.

PR #87, opened 2020-08-22
Improve logs

### LINKED ISSUE #54: Handle what happens if dartfmt is not avaiable on the path
We should at least print a more human-friendly error incase dartfmt is not available.
We can also look into `flutter format` in case someone is using dart via the flutter sdk.

https://github.com/dart-lang/ffigen/issues/26#issuecomment-663829440

mannprerak2:
Closes #54 
- Added fallback to `flutter format` if `dartfmt` is not on path.
- Added ANSI colors for printing logs (red for severe logs, green for success message)


lib/src/code_generator/library.dart
```diff
@@ -93,10 +93,23 @@ class Library {
...
+        final result = Process.runSync('flutter', ['format', path],
+            runInShell: Platform.isWindows);
+        if (result.stderr.toString().isNotEmpty) {
+          _logger.severe(result.stderr);
+        }
+      } on ProcessException {
+        _logger.severe(
+            "Couldn't format bindings, unable to call 'dartfmt' or 'flutter format'.");
+      }
     }
```
jonasfj, line 113: If you want to find something in the current SDK, I would suggest using:
https://pub.dev/documentation/cli_util/latest/cli_util/getSdkPath.html

Then you can always find a dart-sdk..

But you could also just have a dependency on https://pub.dev/packages/dart_style which is the library version.

jonasfj (approved):

PR #106, opened 2020-09-20
Fixed errors due to extended ASCII and control characters in macros

### LINKED ISSUE #105: Unexpected extension byte
I'm trying to generate bindings for the posix libraries.

I'm getting a unhandled exception:

```
I_generated_macro_variable
[FINE]   : macroVariablevisitor: (Cursor) spelling: _17_TIPC_MIN_LINK_TOL_generated_macro_variable, kind: 9, kindSpelling: VarDecl, type: 118, typeSpelling: int, usr: c:@_17_TIPC_MIN_LINK_TOL_generated_macro_variable
[FINE]   : macroVariablevisitor: (Cursor) spelling: _17_TIPC_DEF_LINK_TOL_generated_macro_variable, kind: 9, kindSpelling: VarDecl, type: 118, typeSpelling: int, usr: c:@_17_TIPC_DEF_LINK_TOL_generated_macro_variable
[FINE]   : macroVariablevisitor: (Cursor) spelling: _17_TIPC_MAX_LINK_TOL_generated_macro_variable, kind: 9, kindSpelling: VarDecl, type: 118, typeSpelling: int, usr: c:@_17_TIPC_MAX_LINK_TOL_generated_macro_variable
[FINE]   : macroVariablevisitor: (Cursor) spelling: _17_TIPC_MIN_LINK_WIN_generated_macro_variable, kind: 9, kindSpelling: VarDecl, type: 118, typeSpelling: int, usr: c:@_17_TIPC_MIN_LINK_WIN_generated_macro_variable
[FINE]   : macroVariablevisitor: (Cursor) spelling: _17_TIPC_DEF_LINK_WIN_generated_macro_variable, kind: 9, kindSpelling: VarDecl, type: 118, typeSpelling: int, usr: c:@_17_TIPC_DEF_LINK_WIN_generated_macro_variable
[FINE]   : macroVariablevisitor: (Cursor) spelling: _17_TIPC_MAX_LINK_WIN_generated_macro_variable, kind: 9, kindSpelling: VarDecl, type: 118, typeSpelling: int, usr: c:@_17_TIPC_MAX_LINK_WIN_generated_macro_variable
[FINE]   : macroVariablevisitor: (Cursor) spelling: _21_TIPC_DEF_LINK_UDP_MTU_generated_macro_variable, kind: 9, kindSpelling: VarDecl, type: 118, typeSpelling: int, usr: c:@_21_TIPC_DEF_LINK_UDP_MTU_generated_macro_variable
[FINE]   : macroVariablevisitor: (Cursor) spelling: _17_TIPC_NTQ_ALLTYPES_generated_macro_variable, kind: 9, kindSpelling: VarDecl, type: 118, typeSpelling: unsigned int, usr: c:@_17_TIPC_NTQ_ALLTYPES_generated_macro_variable
[FINE]   : macroVariablevisitor: (Cursor) spelling: _18_TIPC_CFG_TLV_ERROR_generated_macro_variable, kind: 9, kindSpelling: VarDecl, type: 118, typeSpelling: const char *, usr: c:@_18_TIPC_CFG_TLV_ERROR_generated_macro_variable
[SEVERE] : FormatException: Unexpected extension byte (at offset 0)
[SEVERE] : #0      _Utf8Decoder.convertSingle (dart:convert-patch/convert_patch.dart:1783:7)
#1      Utf8Decoder.convert (dart:convert/utf.dart:321:42)
#2      Utf8Codec.decode (dart:convert/utf.dart:62:20)
#3      Utf8.fromUtf8 (package:ffi/src/utf8.dart:43:17)
#4      _macroVariablevisitor (package:ffigen/src/header_parser/sub_parsers/macro_parser.dart:134:28)
#5      _FfiCallback_macroVariablevisitor (dart:ffi)
#6      FfiTrampoline (dart:ffi)
#7      Clang.clang_visitChildren_wrap (package:ffigen/src/header_parser/clang_bindings/clang_bindings.dart:498:37)
#8      parseSavedMacros (package:ffigen/src/header_parser/sub_parsers/macro_parser.dart:81:30)
#9      parseToBindings (package:ffigen/src/header_parser/parser.dart:128:19)
#10     parse (package:ffigen/src/header_parser/parser.dart:26:20)
#11     main (file:///home/bsutton/.pub-cache/hosted/pub.dartlang.org/ffigen-1.0.2/bin/ffigen.dart:54:19)
#12     _startIsolate.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:299:32)
#13     _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:168:12)

Unhandled exception:
Exception: Exception thrown in a dart function called via C, use --verbose to see more details
#0      visitChildrenResultChecker (package:ffigen/src/header_parser/utils.dart:21:5)
#1      parseSavedMacros (package:ffigen/src/header_parser/sub_parsers/macro_parser.dart:88:5)
#2      parseToBindings (package:ffigen/src/header_parser/parser.dart:128:19)
#3      parse (package:ffigen/src/header_parser/parser.dart:26:20)
#4      main (file:///home/bsutton/.pub-cache/hosted/pub.dartlang.org/ffigen-1.0.2/bin/ffigen.dart:54:19)
#5      _startIsolate.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:299:32)
#6      _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:168:12)
```

My configuration is:
```
name: 'posix'
descriptions: 'Bindings to all posix apis'

output: 'generated_bindings.dart'

headers: 
  entry-points:
      - '/usr/include/linux/**.h'

array-workaround: true      

comments:
    style: doxygen
    length: full

use-supported-typedefs: true  

preamble: |
  /// AUTO GENERATED FILE, DO NOT EDIT.
  ///
  /// Generated by `package:ffigen`.
```

mannprerak2:
Closes #105 
- Added tests, updated version, and changelog

### DIFF
--- a/lib/src/header_parser/sub_parsers/macro_parser.dart
+++ b/lib/src/header_parser/sub_parsers/macro_parser.dart
@@ -132,41 +132,16 @@ int _macroVariablevisitor(Pointer<clang_types.CXCursor> cursor,
           );
           break;
         case clang_types.CXEvalResultKind.CXEval_StrLiteral:
-          final strPtr = clang.clang_EvalResult_getAsStr(e); // Do not free.
-          final length = Utf8.strlen(strPtr.cast());
-          final charList = Uint8List.view(
-              strPtr.cast<Uint8>().asTypedList(length).buffer, 0, length);
-
-          final sb = StringBuffer();
-          for (final char in charList) {
-            if (char >= 32 && char <= 126) {
-              // Handle the printable range of ASCII.
-              final printableChar = String.fromCharCode(char);
-              // Escape $ and ' character.
-              if (printableChar == '\$' ||
-                  printableChar == "'" ||
-                  printableChar == '\\') {
-                sb.write('\\$printableChar');
-              } else {
-                sb.write(printableChar);
-              }
-            } else {
-              /// Range [127..255] are extended ASCII and [0..31, 127] are
-              /// Control characters. Extended ASCII set cannot be decoded by
-              /// dart:convert and the Control characters cannot be printed
-              /// either, so we use their hexadecimal to write them.
-              /// E.g the New Line character `\n` (ASCII - 10) will be written
-              /// as `\x0A`.
-              final hexa = char.toRadixString(16).toUpperCase().padLeft(2, '0');
-              sb.write('\\x${hexa}');
-            }
-          }
+          final rawValue = _getWrittenRepresentation(
+            macroName,
+            clang.clang_EvalResult_getAsStr(e),
+          );
           constant = Constant(
             usr: savedMacros[macroName].usr,
             originalName: savedMacros[macroName].originalName,
             name: macroName,
             rawType: 'String',
-            rawValue: "'${sb.toString()}'",
+            rawValue: "'${rawValue}'",
           );
           break;
       }
@@ -265,4 +240,95 @@ class MacroVariableString {
     final nameStart = lengthEnd + 1;
     return s.substring(nameStart, nameStart + len);
   }
+}
+
+/// Gets a written representation string of a C string.
+///
+/// E.g- For a string "Hello\nWorld", The new line character is converted to \n.
+/// Note: The string is considered to be Utf8, but is treated as Extended ASCII,
+/// if the conversion fails.
+String _getWrittenRepresentation(String macroName, Pointer<Int8> strPtr) {
+  final sb = StringBuffer();
+  try {
+    // Consider string to be Utf8 encoded by default.
+    sb.clear();
+    // This throws a Format Exception if string isn't Utf8 so that we handle it
+    // in the catch block.
+    final result = Utf8.fromUtf8(strPtr.cast());
+    for (final s in result.runes) {
+      // Handle ASCII control characters separately.
+      if (s >= 0 && s < 32 || s == 127) {
+        /// Handle these - `\b \t \n \v \f \r` as special cases.
+        switch (s) {
+          case 8: // \b
+            sb.write(r'\b');
+            break;
+          case 9: // \t
+            sb.write(r'\t');
+            break;
+          case 10: // \n
+            sb.write(r'\n');
+            break;
+          case 11: // \v
+            sb.write(r'\v');
+            break;
+          case 12: // \f
+            sb.write(r'\f');
+            break;
+          case 13: // \r
+            sb.write(r'\r');
+            break;
+          default: // For any other control character.
+            final h = s.toRadixString(16).toUpperCase().padLeft(2, '0');
+            sb.write('\\x${h}');
+        }
+      } else {
+        sb.write(_getWritableChar(s));
+      }
+    }
+  } catch (e) {
+    // Handle string if it isn't Utf8. String is considered to be
+    // Extended ASCII in this case.
+    _logger.warning(
+        "Couldn't decode Macro string '$macroName' as Utf8, using ASCII instead.");
+    sb.clear();
+    final length = Utf8.strlen(strPtr.cast());
+    final charList = Uint8List.view(
+        strPtr.cast<Uint8>().asTypedList(length).buffer, 0, length);
+
+    for (final char in charList) {
+      if (char >= 32 && char <= 126) {
+        // Handle the printable range of ASCII.
+        final printableChar = _getWritableChar(char);
+        sb.write(printableChar);
+      } else {
+        /// Range [127..255] are extended ASCII and [0..31, 127] are
+        /// Control characters. Extended ASCII set cannot be decoded by
+        /// dart:convert and the Control characters cannot be printed
+        /// either, so we use their hexadecimal to write them.
+        /// E.g the New Line character `\n` (ASCII - 10) will be written
+        /// as `\x0A`.
+        final hexa = char.toRadixString(16).toUpperCase().padLeft(2, '0');
+        sb.write('\\x${hexa}');
+      }
+    }
+  }
+
+  return sb.toString();
+}
+
+/// Creates a writable char from [char] code.
+///
+/// E.g- `\` is converted to `\\`.
+String _getWritableChar(int char) {
+  final res = String.fromCharCode(char);
+  switch (res) {
+    case r"'":
+      return r"\'";
+    case r'$':
+      return r'\$';
+    case r'\':
+      return r'\\';
+  }
+  return res;
 }
--- a/lib/src/header_parser/sub_parsers/macro_parser.dart
+++ b/lib/src/header_parser/sub_parsers/macro_parser.dart
@@ -312,7 +312,7 @@ String _getWritableChar(int char, {bool utf8 = true}) {
       return r'\\';
   }
 
-  /// Incase encoding is not Utf8, we know all characters will fall in [0..255]
+  /// In case encoding is not Utf8, we know all characters will fall in [0..255]
   /// Print range [128..255] as `\xHH`.
   if (!utf8) {
     final h = char.toRadixString(16).toUpperCase().padLeft(2, '0');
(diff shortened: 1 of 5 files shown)


mannprerak2:
@dcharkes 
The current code will represent `\n` in a macro with `\x0A` in the generated Dart constant. Do you think we should change this to use `\n` instead?
If yes, what all characters should we apply it to? (`\n \t \r` are probably the most common)

There are some common [escape sequences in C ](https://en.wikipedia.org/wiki/Escape_sequences_in_C#Table_of_escape_sequences:~:text=The%20following%20escape%20sequences%20are%20defined,a%20character%20encoding%20based%20on%20ASCII.), but I am not sure if all of these are supported in Dart.

lib/src/header_parser/sub_parsers/macro_parser.dart
```diff
@@ -131,17 +132,41 @@ int _macroVariablevisitor(Pointer<clang_types.CXCursor> cursor,
...
+                  printableChar == "'" ||
+                  printableChar == '\\') {
+                sb.write('\\$printableChar');
+              } else {
+                sb.write(printableChar);
+              }
+            } else {
+              /// Range [127..255] are extended ASCII and [0..31, 127] are
+              /// Control characters. Extended ASCII set cannot be decoded by
+              /// dart:convert and the Control characters cannot be printed
```
dcharkes, line 156: `dart:convert` has https://api.dart.dev/stable/2.9.3/dart-convert/AsciiDecoder-class.html a. Does the AsciiDecoder not work for extended ascii?

dcharkes (commented):
Does the `dart:convert` AsciiEncoder/AsciiDecoder not support https://en.wikipedia.org/wiki/Extended_ASCII?

If not, do we know which for of extended ascii clang is giving us through the API?

lib/src/header_parser/sub_parsers/macro_parser.dart
mannprerak2: It does not, unless i do `AsciiDecoder(allowInvalid: true);`.
Even so, because we need to write the characters to a file to be used, we cannot simply print it.
E.g `\x80` is printed as  (it seems to be showing a question mark in GitHub preview) which I suppose isn't something we write as it is in out generated file

mannprerak2:
```
FormatException: Invalid value in input: 128
[SEVERE] : #0      _UnicodeSubsetDecoder.convert (dart:convert/ascii.dart:178:11)
```

mannprerak2: > do we know which for of extended ascii clang is giving us through the API

I am not sure what you mean by this, Clang is simply returning a char* which we need to work with now. `Utf8.toUtf8` from `package:ffi` is also crashing, I suppose it should be handled as well.

dcharkes: `char*` in C could be any encoding, it could be ascii, extended ascii, utf8. Once you try to decode with the wrong expected encoded you get failures as the ones you mentioned. See https://en.wikipedia.org/wiki/Mojibake for the source of replacement characters.

The way to prevent decoding errors is to know what encoding was used to create the `char*` (the list of bytes).

mannprerak2: Oh, I get it.

So for this file (1)
```C
#define TEST "this is a string"
```
The file we generate for parsing this is (2)
```cpp
auto _4_TEST_generated_macro_variable = TEST;
```
and clang returns the value of this variable to us.

Now the question is, does the encoding depend on the file encoding where `TEST` is declared (1)? or does it depend on the encoding used for the file we generated (2)?

lib/src/header_parser/sub_parsers/macro_parser.dart
```diff
@@ -131,17 +132,41 @@ int _macroVariablevisitor(Pointer<clang_types.CXCursor> cursor,
...
+          final sb = StringBuffer();
+          for (final char in charList) {
+            if (char >= 32 && char <= 126) {
+              // Handle the printable range of ASCII.
+              final printableChar = String.fromCharCode(char);
+              // Escape $ and ' character.
+              if (printableChar == '\$' ||
+                  printableChar == "'" ||
+                  printableChar == '\\') {
+                sb.write('\\$printableChar');
```
dcharkes, line 149: You could possibly use raw strings instead to prevent having to escape the chars: `r'This will end with a newline.\n'`

dcharkes:
> @dcharkes
> The current PR will represent `\n` in a macro with `\x0A` in the generated Dart constant. Do you think we should change this to use `\n` instead?
> If yes, what all characters should we apply it to? (`\n \t \r` are probably the most common)
> 
> There are some common [escape sequences in C ](https://en.wikipedia.org/wiki/Escape_sequences_in_C#Table_of_escape_sequences:~:text=The%20following%20escape%20sequences%20are%20defined,a%20character%20encoding%20based%20on%20ASCII.), but I am not sure if all of these are supported in Dart.

How come the current PR generates this? Does clang give us a string with 4 characters? `r'\x0A'`. Or do we get a single character?

Can we solve it with a generic solution? With properly using encoders/decoders? If not, then we might have to special case the C escape sequences.

mannprerak2:
> Does clang give us a string with 4 characters? r'\x0A'. Or do we get a single character?

We get a single character, (E.g for `\n`, and then we convert `128` to its HEX value `80`.

Also, the encoding seems to be dependent on the string itself.
This `#define TIPC_CFG_TLV_ERROR "😄"` works perfectly with `Utf8.fromUtf8`, but doesn't work with current PR because we considered it as extended ASCII (and so `'\xF0\x9F\x98\x84'` is generated.)
But `#define TIPC_CFG_TLV_ERROR "\x80" doesn't work with `Utf8.fromUtf8` because clang returns us a single character (Extended ASCII - 128) for this.

> Can we solve it with a generic solution? With properly using encoders/decoders?

We'd probably need one for Extended ASCII, but I am not sure how to determine the encoding used, because extended ASCII collides with UTF8

mannprerak2:
We have to do the following -
1. Treat everything as Utf8 and if Utf8 decoding fails, we print the printable ASCII as it is, and use `\xHH` for control characters and extended ASCII. This will prevent any crashes.

2. We can give the user a way to specify which encoding to use., We can add another pattern in the config for specifying the encoding to use (default is `1`), and the user can match the macro name with a regexp.

For `1`, if the string is not Utf8 but we are able to decode it with Utf8, we will end up with some garbled text in Dart. But at least there shouldn't be any problem when converting that string back to `char*`, as the bytes will not change. (In these cases, the user will need to specify the encoding to use for that particular macro using `2`.)

Perhaps we can cover `2` in another PR. What do you think @dcharkes?

lib/src/header_parser/sub_parsers/macro_parser.dart
mannprerak2: Yes I suppose we can, this will handle `$` and `\` but we'd still need to escape `'`.

dcharkes:
Sounds good 👍

lib/src/header_parser/sub_parsers/macro_parser.dart
mannprerak2: > Yes I suppose we can

We can't actually, because we may have to write things like `\x0A` to our string and also escape `'`.

lib/src/header_parser/sub_parsers/macro_parser.dart
```diff
@@ -241,3 +241,84 @@ class MacroVariableString {
...
+  switch (char) {
+    case 36: // $
+      return r'\$';
+    case 39: // '
+      return r"\'";
+    case 92: // \
+      return r'\\';
+  }
+
+  /// Incase encoding is not Utf8, we know all characters will fall in [0..255]
```
dcharkes, line 315: In case

dcharkes (approved):

mannprerak2:
@dcharkes Fixed
> We can give the user a way to specify which encoding to use., We can add anot...

I have opened a new issue #107 for this.

PR #113, opened 2020-10-10
Handled proper conversion of macro doubles.

### LINKED ISSUE #112: Header files including math.h generate erroneous bindings
Consider a simple `ffigen-config.yaml`:
```yaml
output: 'file.dart'
headers:
  entry-points:
    - './test.h'
```

And the `test.h` file which is just a single include directive:
```c
#include <math.h>
```

This generates a large bindings file, but one part of it seems to violate the Dart analyzer:
```dart
const double HUGE_VAL = Infinity;

const double HUGE_VALF = Infinity;

const double HUGE_VALL = Infinity;

const double INFINITY = Infinity;

const double NAN = NaN;
```

`Infinity` and `NaN` are not valid Dart constants, and instead should be `double.infinity` and `double.nan` respectively.

mannprerak2:
Closes #112 
- Handled conversion of macro doubles properly (`double.Infinity`, `double.negativeInfinity` and `double.nan`).
- Updated version and changelog.
- Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 1.0.5
+- Fixed issues with generating macros of type `double.Infinity` and `double.NaN`.
+
 # 1.0.4
 - Updated code to use `dart format` instead of `dartfmt` for sdk version `>= 2.10.0`.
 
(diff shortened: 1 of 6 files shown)


dcharkes (approved):
lgtm

PR #121, opened 2020-11-20
Added support for Dart_Handle to be generated as Handle

### LINKED ISSUE #120: Support Dart_Handle from dart_api.h
In https://github.com/dart-lang/sdk/commit/6544c69e23256f7d60498e6f5f66a395a008587f we added support for passing any Dart `Object` to native as a `Dart_Handle` from the [dart_api.h](https://github.com/dart-lang/sdk/blob/master/runtime/include/dart_api.h).

It would be nice if we could use `package:ffigen` to automatically generate bindings for `Dart_Handle`.

* c type to recognise: `Dart_Handle`
* type marker to generate for C type signatures in Dart: `Handle`
* Dart type to generate: `Object`

For more info see the CL and the tests/examples added in the CL.

mannprerak2:
Closes #120 
- Dart_Handle is now mapped directly to Handle/Object.
- Added config option `use-dart-handle` to enable(default) or disable generating Handles.
- Updated readme, version and changelog.

---

I am comparing the USR for the underlying struct to determine if it's `Dart_Handle`.
```dart
const dartHandleUsr = 'c:@S@_Dart_Handle';
```

This **can** be confused with a user-defined Struct named `_Dart_Handle`. So in case, the user has a struct called `_Dart_handle` they can disable Handle generation using `use-dart-handle: false`.

Also, I am hoping the name of the underlying struct (`_Dart_Handle` as defined in `dart_api.h`) won't change or this feature will break.
```c
typedef struct _Dart_Handle* Dart_Handle;
```

If a struct has `Dart_Handle`, all its members will be removed. However using `Dart_Handle*` will work and be generated as `Pointer<Handle>`.
```c
// Dart_Handle isn't supported directly, so all members are removed.
struct struc1
{
    Dart_Handle h;
    int a;
};

// Pointer<Handle> works.
struct struc2
{
    Dart_Handle *h;
};
```
### Sample I/O

```C
Dart_Handle func(Dart_Handle **a);
```

```dart
class NativeLibrary {
  /// Holds the Dynamic library.
  final ffi.DynamicLibrary _dylib;

  NativeLibrary(ffi.DynamicLibrary dynamicLibrary) : _dylib = dynamicLibrary;

  Object func(
    ffi.Pointer<ffi.Pointer<ffi.Handle>> a,
  ) {
    _func ??= _dylib.lookupFunction<_c_func, _dart_func>('func');
    return _func(
      a,
    );
  }

  _dart_func _func;
}

typedef _c_func = ffi.Handle Function(
  ffi.Pointer<ffi.Pointer<ffi.Handle>> a,
);

typedef _dart_func = Object Function(
  ffi.Pointer<ffi.Pointer<ffi.Handle>> a,
);
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 1.2.0
+- Added support for `Dart_Handle` from `dart_api.h`.
+
 # 1.1.0
 - `typedef-map` can now be used to map a typedef name to a native type directly.
 
--- a/README.md
+++ b/README.md
@@ -230,7 +230,19 @@ dart-bool: true
 ```
   </td>
   </tr>
-   <tr>
+  <tr>
+    <td>use-dart-handle</td>
+    <td>Should map `Dart_Handle` to `Handle`.<br>
+    <b>Default: true</b>
+    </td>
+    <td>
+
+```yaml
+use-dart-handle: true
+```
+  </td>
+  </tr>
+  <tr>
     <td>preamble</td>
     <td>Raw header of the file, pasted as-it-is.</td>
     <td>
--- a/lib/src/code_generator/type.dart
+++ b/lib/src/code_generator/type.dart
@@ -39,6 +39,9 @@ enum BroadType {
   Struct,
   NativeFunction,
 
+  /// Represents a Dart_Handle.
+  Handle,
+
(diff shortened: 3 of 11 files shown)


dcharkes (approved):
Cool!

PR #122, opened 2020-12-18
Null safety migration

mannprerak2:
Closes #8 

- Migrated code for null safety (unsound null safety).
- Struct members are now marked as `external`.
- Updated readme, version, changelog, examples, and tests.

---

- [x] Migrating code to pass the null safe syntax
- [x] Optimising code properly to take full advantage of null safety
- [x] Generating null safe code

PR #126, opened 2021-01-02
Support nested structs

### LINKED ISSUE #4: Support for Nested Structures
Waiting on support for [Nested Structs](https://github.com/dart-lang/sdk/issues/37271) in `dart:ffi`
Currently, All members of a Struct are removed from bindings if it has another struct as a member

jpnurmi:
Fixes #4

### DIFF
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -14,13 +14,13 @@ env:
 
 jobs:
   # Check code formatting and static analysis on a single OS (linux)
-  # against Dart stable.
+  # against Dart master.
   analyze:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        sdk: [stable]
+        sdk: [master]
     steps:
       - uses: actions/checkout@v2
       - uses: dart-lang/setup-dart@v0.1
@@ -44,7 +44,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: dart-lang/setup-dart@v0.1
         with:
-          channel: stable
+          channel: master
       - name: Install dependencies
         run: dart pub get  
       - name: Install libclang-10-dev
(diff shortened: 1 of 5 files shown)


mannprerak2:
We can also target #3 along with this.

And It is better to put these behind a flag in config for users to enable this, since this will only work on dart master branch as of now.

I think this should be held off until we merge #122 .

dcharkes:
Super cool stuff! 👍 

I second @mannprerak2, we should not merge this into the master branch until nested structs also becomes available in Dart stable. If we merge it into master we can no longer make any changes to `package:ffigen` and release them for Dart stable in the mean time.
(We could make a new branch `dart-master` which we build against the master branch of Dart as you did in this PR in `test-package.yml`. And release a `1.3.0-nested-structs.0` version. However, you can also just use a git dependency to depend on your own branch for the time being. Any preference @jpnurmi ?)

@mannprerak2 do you want to do the reviewing for this PR?

jpnurmi:
Sure, I was planning to get back to this after #122 has been merged. This can be kept in an experimental branch or made configurable as @mannprerak2 suggested. There's no rush from my side to have this upstream, I got the files already generated. :)

mannprerak2:
> @mannprerak2 do you want to do the reviewing for this PR?

Sure.

mannprerak2:
@jpnurmi I think it's best if we create an experimental branch for this (`dart-master` as Daco suggested). Otherwise, we 'dneed to add and then later remove those configs, modify tests to run according to certain configurations. 

Also, we can target #3 as a separate PR, since it needs to know if a struct had been properly generated to be able to pass it by value correctly.

mannprerak2:
@dcharkes I think you would need to create a dart-master branch so that @jpnurmi can make a PR on it.

dcharkes:
Done: https://github.com/dart-lang/ffigen/tree/dart-master

PR #134, opened 2021-01-08
Support for passing/returning Structs by value.

### LINKED ISSUE #3: Support for Passing/Returning Structs by value in Functions
Waiting on support for [Structs by value](https://github.com/dart-lang/sdk/issues/41062) in `dart:ffi`
Currently, functions passing/returning Structs by value are removed from the generated bindings

mannprerak2:
Closes #3 
- Added support for passing/returning structs by value.
- Added, updated tests, examples.
- Updated changelog, readme, version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 2.0.0-dev.1
+- Added support for passing and returning struct by value in functions.
+
 # 2.0.0-dev.0
 - Added support for Nested structs.
 
--- a/README.md
+++ b/README.md
@@ -377,8 +377,6 @@ class ArrayHelper_CXFileUniqueID_data_level0 {
 
 ## Limitations
 1. Multi OS support for types such as long. [Issue #7](https://github.com/dart-lang/ffigen/issues/7)
-2. Function's passing/returning structs by value are skipped. [Issue #3](https://github.com/dart-lang/ffigen/issues/3)
-3. Structs containing structs will have all their members removed. [Issue #4](https://github.com/dart-lang/ffigen/issues/4)
 
 ## Trying out examples
 1. `cd examples/<example_u_want_to_run>`, Run `pub get`.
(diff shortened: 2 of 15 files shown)


mannprerak2:
@dcharkes Hey, I wanted to know what will happen if we pass a struct (with an incomplete definition by value), but the function was not really expecting that struct?

Currently, I am making sure to not generate any functions which pass/return these incomplete structs by value. Should I be doing this?

dcharkes:
I'm assuming C would give a static error on trying to do something with an incomplete type? If that is the case, then not generating the functions and emitting a warning/error is the right approach.

dcharkes (commented):
We should be able to get rid of `lib/src/clang_library/wrapper.c` in this CL right?

mannprerak2:
> We should be able to get rid of lib/src/clang_library/wrapper.c in this CL right?

We can but I think we should do that in another PR.

dcharkes (approved):
This is a smaller PR than I would have expected. 👍 

lgtm

You might want to make the next PR which removes c wrapper just to dogfood test this PR before we merge it.

mannprerak2:
> I'm assuming C would give a static error on trying to do something with an incomplete type? If that is the case, then not generating the functions and emitting a warning/error is the right approach.

I meant, that sometimes when we remove all struct members if any of the members are unsupported (like bitfield).
So the struct we generate is empty.  So in that case when a user passes this struct by value, how does the dart VM act on this?

mannprerak2:
> You might want to make the next PR which removes C wrapper just to dogfood test this PR before we merge it.

Okay sure

dcharkes:
> > I'm assuming C would give a static error on trying to do something with an incomplete type? If that is the case, then not generating the functions and emitting a warning/error is the right approach.
> 
> I meant, that sometimes when we remove all struct members if any of the members are unsupported (like bitfield).
> So the struct we generate is empty. So in that case when a user passes this struct by value, how does the dart VM act on this?

That would result in segfaults because we would be talking about different registers/stack values. However, empty structs are undefined behavior on C, so `dart:ffi` rejects empty structs in `asFunction` and `fromFunction`. So if you would continue generating, you would generate code that would then be rejected by the Dart compiler.

dcharkes (approved):

PR #135, opened 2021-01-08
Remove setup phase

### LINKED ISSUE #61: Removing wrapper.c
We currently need `wrapper.c` because `dart:ffi` does not support passing structs by value yet (https://github.com/dart-lang/ffigen/issues/3).

However, we also have other things in `wrapper.c`, for example having some process uuid and a stack.

We should investigate whether we can remove those things and do it in Dart. Or document why it is not possible to do it in only Dart + `dart:ffi`. (That might tell us we need more features in `dart:ffi`.)

Also, we could make a branch for this package which works on the locally built https://dart-review.googlesource.com/c/sdk/+/140290, so that we can merge that brach after structs-by-value lands and is released in stable.

mannprerak2:
Closes #61 
- [x] Remove the usage of `_wrap` functions from code.
- [x] Remove wrapper.c and setup phase for ffigen.
---

Follow up for #134.


mannprerak2:
@dcharkes This helped remove 1 bug from #134, I think we can merge that now.

For task2 (removing setup phase) we need to have the user input the location of libclang's dynamic library. I am thinking we can have a few default locations to check, but if we can't find it there, we make the user enter the `path/to/llvm/lib` so that we can find the dynamic library there.

mannprerak2:
This is complete now, I've added a new config key - `llvm-lib` which is required if we can't find it at the default locations.

dcharkes:
@mannprerak2 this is marked as draft, do you want me to start reviewing?

And do you want to merge this in before releasing `2.0.0-dev.1`?

mannprerak2:
> do you want me to start reviewing?

Yes, I have added a merge commit with dart-dev branch, so that should be easier now.

lib/src/README.md
```diff
@@ -38,19 +21,11 @@ This is the main entry point for the user-  `pub run ffigen`.
 - Command-line options:
     - `--verbose`: Sets log level.
     - `--config`: Specifies a config file.
-- `ffigen.dart` will first check if a dynamic library already exists and is up to date. If not, it tries to auto-create it. If that fails, user must excplicitly call [setup.dart](#setup.dart).
+- `ffigen.dart` will first try to find dynamic library in default locations. If not. If that fails, user must excplicitly specify location in config under the key `llvm-lib`.
```
dcharkes, line 24: "If not." is not a sentence, it does not have a verb.

dcharkes (approved):
Cool! Really cool to see that wrapper gone! 🥳

This shows structs by value is a nice `dart:ffi` feature to have! 😃

lib/src/README.md
mannprerak2: Ah, seems like I left that out by mistake. Fixed it.

>This shows structs by value is a nice dart:ffi feature to have! 

Absolutely 🎉

Issue #128, opened 2021-01-05
Unsound Null-Safety


mannprerak2:
We currently do not support sound null safety (added by #122).

Tasks Remaining: 
- [x] Migrate all dependencies
  - [x] Args
  - [x] logging
  - [x] cli_util
  - [x] glob
  - [x] pub_semver
- [x] Remove all `--no-sound-null-safety` flags.

TimWhiting:
I'm having issues trying to do this.
```
$ dart --no-sound-null-safety pub global activate ffigen 2.0.0-dev.2
Package ffigen is currently active at version 1.2.0.
Resolving dependencies... (1.7s)
+ args 1.6.0
+ async 2.4.2
+ charcode 1.2.0-nullsafety.3
+ cli_util 0.2.0
+ collection 1.15.0-nullsafety.5
+ ffi 0.2.0-nullsafety.1
+ ffigen 2.0.0-dev.2 (2.0.0-nullsafety.0 available)
+ file 5.2.1
+ glob 1.2.0
+ intl 0.16.1
+ js 0.6.2
+ logging 0.11.4
+ matcher 0.12.10-nullsafety.3
+ meta 1.3.0-nullsafety.6
+ node_interop 1.2.1
+ node_io 1.2.0
+ path 1.8.0-nullsafety.3
+ pedantic 1.9.2
+ pub_semver 1.4.4
+ quiver 3.0.0-nullsafety.2
+ source_span 1.8.0-nullsafety.4
+ stack_trace 1.10.0-nullsafety.6
+ string_scanner 1.1.0-nullsafety.3
+ term_glyph 1.2.0-nullsafety.3
+ yaml 3.0.0-nullsafety.0
Precompiling executables... (1.0s)
Failed to precompile ffigen:ffigen:
Error: Cannot run with sound null safety, because the following dependencies
don't support null safety:

 - package:args
 - package:cli_util
 - package:logging
 - package:pub_semver
 - package:glob
 - package:async
 - package:pedantic

For solutions, see https://dart.dev/go/unsound-null-safety
```

By having the correct version in dev dependencies and running it in the project I get a similar thing:

```
$ dart --no-sound-null-safety run ffigen
Precompiling executable... (1.1s)
Failed to precompile ffigen:ffigen:
Error: Cannot run with sound null safety, because the following dependencies
don't support null safety:

 - package:args
 - package:cli_util
 - package:logging
 - package:pub_semver
 - package:glob
 - package:async
[...]

PR #136, opened 2021-01-10
Removed usage of --no-sound-null-safety flag.

mannprerak2:
The `--no-sound-null-safety` flag was causing an issue when package was being run, since it is first precompiled. See comments on #128.
This PR fixes that, it follows the approach of the package:test => https://github.com/dart-lang/test/blob/master/pkgs/test/bin/test.dart.

PR #137, opened 2021-01-10
[dev] Removed the usage of --no-sound-null-safety

mannprerak2:
The --no-sound-null-safety flag was causing an issue when package was being run, since it is first precompiled. See comments on #128.
This PR fixes that **for the dart-dev branch**, it follows the approach of the package:test => https://github.com/dart-lang/test/blob/master/pkgs/test/bin/test.dart.


dcharkes:
It's probably better to merge in master into this PR after merging https://github.com/dart-lang/ffigen/pull/136 to master, and only update pubspec.yaml / readme / changelog.

mannprerak2:
@dcharkes you can review this now.

dcharkes (approved):

PR #139, opened 2021-01-12
Globals

### LINKED ISSUE #138: Support Globals
I'm trying to create a dart wrapper to the Python C API, so that I can call python code from dart easily / pass objects back and forth. 

I was using a different library to generate the ffi bindings, but am now trying to switch to this library. One issue that I have is that there are some global objects that are present in the dynamic / shared library / header files, but not in the generated bindings.

See for example the _Py_TrueStruct, and _Py_FalseStruct [here](https://github.com/python/cpython/blob/master/Include/boolobject.h)

My current workaround and generated bindings are [here](https://github.com/TimWhiting/dartpy/tree/ffigen/dartpy/lib/src/ffi). Essentially just adding another file and extensions on the generated class and reexporting everything. I'm not sure how many definitions I might be missing by just manually adding globals though.

As a side note there are a few macros that would be useful if they could be generated into functions. I know they couldn't be strongly typed, but since dart has a dynamic, it seems plausible, though probably fairly difficult since C macros are essentially copy / paste style macros, and might not necessarily be valid function definitions.

TimWhiting:
Closes #138

@mannprerak2 
I think I made the needed changes. Tested it locally and it seems to work.

### DIFF
--- a/lib/src/code_generator/global.dart
+++ b/lib/src/code_generator/global.dart
@@ -64,7 +64,7 @@ class Global extends LookUpBinding {
     final refOrValue = type.broadType == BroadType.Struct ? 'ref' : 'value';
 
     s.write(
-        "late final ffi.Pointer<$dartType> $pointerName = ${w.dylibIdentifier}.lookup<$cType>('$originalName');\n\n");
+        "late final ${w.ffiLibraryPrefix}.Pointer<$dartType> $pointerName = ${w.dylibIdentifier}.lookup<$cType>('$originalName');\n\n");
     s.write('$dartType get $globalVarName => $pointerName.$refOrValue;\n\n');
     if (type.broadType != BroadType.Struct) {
       s.write(
--- a/lib/src/code_generator/global.dart
+++ b/lib/src/code_generator/global.dart
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'package:ffigen/src/code_generator/typedef.dart';
+
 import 'binding.dart';
 import 'binding_string.dart';
 import 'type.dart';
@@ -33,6 +35,22 @@ class Global extends LookUpBinding {
           name: name,
           dartDoc: dartDoc,
         );
+
+  List<Typedef>? _typedefDependencies;
+  @override
+  List<Typedef> getTypedefDependencies(Writer w) {
+    if (_typedefDependencies == null) {
+      _typedefDependencies = <Typedef>[];
+
+      // Add typedef's required by return type.
+      final returnTypeBase = type.getBaseType();
+      if (returnTypeBase.broadType == BroadType.NativeFunction) {
+        _typedefDependencies!
+            .addAll(returnTypeBase.nativeFunc!.getDependencies());
+      }
+    }
+    return _typedefDependencies!;
+  }
 
   @override
   BindingString toBindingString(Writer w) {
--- a/lib/src/code_generator/global.dart
+++ b/lib/src/code_generator/global.dart
@@ -44,10 +44,11 @@ class Global extends LookUpBinding {
 
     final holderVarName =
         w.wrapperLevelUniqueNamer.makeUnique('_$globalVarName');
+    final refOrValue = type.broadType == BroadType.Struct ? 'ref' : 'value';
     s.write(
-        '${w.ffiLibraryPrefix}.Pointer<${type.getCType(w)}> $holderVarName;\n');
-    s.write(
-        "${type.getDartType(w)} get $globalVarName => ($holderVarName ??= ${w.dylibIdentifier}.lookup<${type.getCType(w)}>('$originalName')).value;\n\n");
+        "late final ${type.getDartType(w)} $holderVarName = ${w.dylibIdentifier}.lookup<${type.getCType(w)}>('$originalName').$refOrValue;\n\n");
+    // s.write(
+    //     "${type.getDartType(w)} get $globalVarName => ($holderVarName ??= ${w.dylibIdentifier}.lookup<${type.getCType(w)}>('$originalName')).value;\n\n");
 
     return BindingString(type: BindingStringType.global, string: s.toString());
   }
--- a/lib/src/code_generator/global.dart
+++ b/lib/src/code_generator/global.dart
@@ -47,8 +47,6 @@ class Global extends LookUpBinding {
     final refOrValue = type.broadType == BroadType.Struct ? 'ref' : 'value';
     s.write(
         "late final ${type.getDartType(w)} $holderVarName = ${w.dylibIdentifier}.lookup<${type.getCType(w)}>('$originalName').$refOrValue;\n\n");
-    // s.write(
-    //     "${type.getDartType(w)} get $globalVarName => ($holderVarName ??= ${w.dylibIdentifier}.lookup<${type.getCType(w)}>('$originalName')).value;\n\n");
 
     return BindingString(type: BindingStringType.global, string: s.toString());
   }
--- a/lib/src/code_generator/global.dart
+++ b/lib/src/code_generator/global.dart
@@ -42,11 +42,10 @@ class Global extends LookUpBinding {
     if (_typedefDependencies == null) {
       _typedefDependencies = <Typedef>[];
 
-      // Add typedef's required by return type.
-      final returnTypeBase = type.getBaseType();
-      if (returnTypeBase.broadType == BroadType.NativeFunction) {
-        _typedefDependencies!
-            .addAll(returnTypeBase.nativeFunc!.getDependencies());
+      // Add typedef's required by the variable's type.
+      final valueType = type.getBaseType();
+      if (valueType.broadType == BroadType.NativeFunction) {
+        _typedefDependencies!.addAll(valueType.nativeFunc!.getDependencies());
       }
     }
     return _typedefDependencies!;
--- a/lib/src/code_generator/global.dart
+++ b/lib/src/code_generator/global.dart
@@ -58,10 +58,18 @@ class Global extends LookUpBinding {
     if (dartDoc != null) {
       s.write(makeDartDoc(dartDoc!));
     }
-
+    final pointerName = w.wrapperLevelUniqueNamer.makeUnique('_$globalVarName');
+    final dartType = type.getDartType(w);
+    final cType = type.getCType(w);
     final refOrValue = type.broadType == BroadType.Struct ? 'ref' : 'value';
+
     s.write(
-        "late final ${type.getDartType(w)} $globalVarName = ${w.dylibIdentifier}.lookup<${type.getCType(w)}>('$originalName').$refOrValue;\n\n");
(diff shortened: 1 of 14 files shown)


TimWhiting:
Okay, this should be ready for review. The only issue right now is the large integration test, which I'm assuming fails because the library maybe has some globals that weren't generated before?

lib/src/header_parser/sub_parsers/var_parser.dart
```diff
@@ -0,0 +1,31 @@
...
+import '../utils.dart';
+
+final _logger = Logger('ffigen.header_parser.var_parser');
+
+/// Parses a global variable
+Global? parseVarDefinition(clang_types.CXCursor cursor) {
+  final type = cursor.type().toCodeGenType();
+  if (type.broadType == BroadType.Unimplemented) {
+    _logger
+        .warning('Global Type not supported $type ${cursor.type().spelling()}');
```
mannprerak2, line 20: Add these logs instead.
```dart
_logger.fine(
          '---- Removed Global, reason: unsupported type: ${cursor.completeStringRepr()}');
_logger.warning(
      "Skipped global variable '$globalName', type not supported.");
```

lib/src/code_generator/global.dart
```diff
@@ -42,12 +59,9 @@ class Global extends LookUpBinding {
...
     }
 
-    final holderVarName =
-        w.wrapperLevelUniqueNamer.makeUnique('_$globalVarName');
-    s.write(
-        '${w.ffiLibraryPrefix}.Pointer<${type.getCType(w)}> $holderVarName;\n');
+    final refOrValue = type.broadType == BroadType.Struct ? 'ref' : 'value';
     s.write(
-        "${type.getDartType(w)} get $globalVarName => ($holderVarName ??= ${w.dylibIdentifier}.lookup<${type.getCType(w)}>('$originalName')).value;\n\n");
+        "late final ${type.getDartType(w)} $globalVarName = ${w.dylibIdentifier}.lookup<${type.getCType(w)}>('$originalName').$refOrValue;\n\n");
```
mannprerak2, line 64: I think global variables are not constants, they can be changed.

If that's the case, we should instead add getters and setters for this.

So for an int global variable, we can do this -
```dart
  late final ffi.Pointer<ffi.Int32> _globalVar =
      _dylib.lookup<ffi.Int32>('globalVar');

  int get globalVar => _globalVar.value;

  set globalVar(int value) => _globalVar.value = value;

```

lib/src/header_parser/translation_unit_parser.dart
```diff
@@ -58,6 +59,9 @@ int _rootCursorVisitor(clang_types.CXCursor cursor, clang_types.CXCursor parent,
         case clang_types.CXCursorKind.CXCursor_MacroDefinition:
           saveMacroDefinition(cursor);
           break;
+        case clang_types.CXCursorKind.CXCursor_VarDecl:
+          addToBindings(parseVarDefinition(cursor));
```
mannprerak2, line 63: I think `parseVarDeclaration` will be a more consistent name for this function.

mannprerak2 (changes requested):
Can you also add a new test in `test/header_parser_tests`? You can pretty much copy the `unnamed_enum_test.dart` file and modify that to add a few tests. (say for a pointer, int, struct, Long Double, pointer to Long double)

mannprerak2:
> The only issue right now is the large integration test, which I'm assuming fails because the library maybe has some globals that weren't generated before?

The `test/debug_generated` folder will contained failed tests, you can compare those with `_expected.*` files  in `large_integration_tests` folder, and subsequently add the changes there if everything seems okay.

lib/src/header_parser/sub_parsers/var_parser.dart
```diff
@@ -0,0 +1,31 @@
...
+import '../clang_bindings/clang_bindings.dart' as clang_types;
+import '../data.dart';
+import '../utils.dart';
+
+final _logger = Logger('ffigen.header_parser.var_parser');
+
+/// Parses a global variable
+Global? parseVarDefinition(clang_types.CXCursor cursor) {
+  final type = cursor.type().toCodeGenType();
+  if (type.broadType == BroadType.Unimplemented) {
```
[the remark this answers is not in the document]
TimWhiting: Done

lib/src/header_parser/sub_parsers/var_parser.dart
TimWhiting: Done

lib/src/code_generator/global.dart
TimWhiting: Yes, this works for all types except structs, which just don't have a setter, but fields for structs can be set from the ref that is obtained via the getter.

TimWhiting:
@mannprerak2 
I addressed everything you mentioned, and added a few features that were missing in my previous implementation (the exclude configuration). Added tests, and updated the _expected file in the large integration test.

lib/src/code_generator/global.dart
```diff
@@ -34,20 +36,40 @@ class Global extends LookUpBinding {
...
+    final cType = type.getCType(w);
+    final refOrValue = type.broadType == BroadType.Struct ? 'ref' : 'value';
 
-    final holderVarName =
-        w.wrapperLevelUniqueNamer.makeUnique('_$globalVarName');
     s.write(
-        '${w.ffiLibraryPrefix}.Pointer<${type.getCType(w)}> $holderVarName;\n');
-    s.write(
-        "${type.getDartType(w)} get $globalVarName => ($holderVarName ??= ${w.dylibIdentifier}.lookup<${type.getCType(w)}>('$originalName')).value;\n\n");
+        "late final ffi.Pointer<$dartType> $pointerName = ${w.dylibIdentifier}.lookup<$cType>('$originalName');\n\n");
```
mannprerak2, line 67: `ffi` => `${w.ffiLibraryPrefix}`, this is needed since this prefix will change if some C library has a declaration named `ffi`.

lib/src/header_parser/sub_parsers/var_parser.dart
```diff
@@ -0,0 +1,39 @@
...
+import '../utils.dart';
+
+final _logger = Logger('ffigen.header_parser.var_parser');
+
+/// Parses a global variable
+Global? parseVarDeclaration(clang_types.CXCursor cursor) {
+  final type = cursor.type().toCodeGenType();
+  if (type.broadType == BroadType.Unimplemented ||
+      type.getBaseType().isIncompleteStruct ||
+      type.getBaseType().unimplementedReason != null) {
```
mannprerak2, line 21: You can simply check the BroadType, no need to check if unimplementedReason is null. 
```dart
if (type.getBaseType().broadType == BroadType.Unimplemented ||
    type.getBaseType().isIncompleteStruct){
...
}
```

lib/src/header_parser/sub_parsers/var_parser.dart
```diff
@@ -0,0 +1,39 @@
...
+        '---- Removed Global, reason: unsupported type: ${cursor.completeStringRepr()}');
+    _logger.warning(
+        "Skipped global variable '${cursor.spelling()}', type not supported.");
+    return null;
+  }
+  if (!shouldIncludeGlobalVar(cursor.usr(), cursor.spelling())) {
+    return null;
+  }
+  final g = Global(
+    originalName: cursor.spelling(),
```
mannprerak2, line 32: You also need to add this global to seen globals, If the global is already seen, you can simply return that.
This is needed since the user can include the same header twice, and when that happens, this would generate 2 bindings (say globalVar and globalVar_1).

 You need to define 2 more functions for Globals in lib/src/header_parser/utils.dart file. Similar to this -
```dart

  void addFuncToSeen(String usr, Func func) {
    _functions[usr] = func;
  }

  Func? getSeenFunc(String usr) {
    return _functions[usr];
  }
```

mannprerak2 (changes requested):
Hi @TimWhiting, Just a few last changes.

lib/src/header_parser/sub_parsers/var_parser.dart
```diff
@@ -0,0 +1,39 @@
...
+  if (type.broadType == BroadType.Unimplemented ||
+      type.getBaseType().isIncompleteStruct ||
+      type.getBaseType().unimplementedReason != null) {
+    _logger.fine(
+        '---- Removed Global, reason: unsupported type: ${cursor.completeStringRepr()}');
+    _logger.warning(
+        "Skipped global variable '${cursor.spelling()}', type not supported.");
+    return null;
+  }
+  if (!shouldIncludeGlobalVar(cursor.usr(), cursor.spelling())) {
```
mannprerak2, line 28: Move this up, so that we don't log about globals which the user has excluded.

mannprerak2 (approved):
lgtm! Thanks @TimWhiting

pubspec.yaml
```diff
@@ -3,23 +3,23 @@
...
 
 name: ffigen
-version: 2.0.0-dev.3
+version: 2.0.0-dev.4
 homepage: https://github.com/dart-lang/ffigen
 description: Experimental generator for FFI bindings, using LibClang to parse C header files.
 
 environment:
-  sdk: '>=2.12.0-198.0.dev <3.0.0'
+  sdk: ">=2.12.0-198.0.dev <3.0.0"
```
dcharkes, line 11: Maybe undo these formatting changes.

If you want to do formatting changes, please make a separate PR and make the GitHub actions check the formatting.

pubspec.yaml
```diff
@@ -3,23 +3,23 @@
...
   ffi: ^0.2.0-nullsafety.1
   yaml: ^3.0.0-nullsafety.0
   path: ^1.8.0-nullsafety.3
   quiver: ^3.0.0-nullsafety.2
-  args: ^1.6.0        # Not available
-  logging: ^0.11.4    # test->coverage->logging
-  cli_util: ^0.2.0    # test->analyzer->cli_util
-  glob: ^1.0.3        # test->analyzer->glob
-  pub_semver: ^1.4.4  # test->analyzer->pub_semver
+  args: ^1.6.0 # Not available
```
dcharkes, line 18: Maybe undo these formatting changes.

If you want to do formatting changes, please make a separate PR and make the GitHub actions check the formatting.

dcharkes (commented):
Cool stuff! Some small comments.

dcharkes:
@TimWhiting 

From the CI:

```
00:13 +40: test/header_parser_tests/globals_test.dart: globals_test (setUpAll)                                                                                                                         
WARNING : Skipped global variable 'longDouble', type not supported.
WARNING : Skipped global variable 'pointerToLongDouble', type not supported.
```

Are those on purpose?

TimWhiting:
> @TimWhiting
> 
> From the CI:
> 
> ```
> 00:13 +40: test/header_parser_tests/globals_test.dart: globals_test (setUpAll)                                                                                                                         
> WARNING : Skipped global variable 'longDouble', type not supported.
> WARNING : Skipped global variable 'pointerToLongDouble', type not supported.
> ```
> 
> Are those on purpose?

Yes, that is on purpose. I test that the globals whose types dart:ffi does not support are not in the generated bindings and those messages are logged so the user knows those variables will not be present in the generated bindings.

PR #142, opened 2021-01-21
Use Opaque to represent empty Structs

### LINKED ISSUE #140: Use the Opaque native type to represent opaque C types
https://github.com/dart-lang/sdk/issues/44622

mannprerak2:
Closes #140 
- Use `Opaque` for representing empty `Struct`s.
- Updated tests, version, changelog
- Fixed Globals, updated Global to directly use `Pointer<EmptyStruct>` instead of `emptyStruct.ref` for getters of `Opaque` types.

---

@dcharkes All the empty structs will now be generated as `Opaque`, this also includes all those structs which we are unable to generate in dart:ffi, which had all their members removed. Is this the correct behavior? or should we handle those structs in some different way.

### DIFF
--- a/test/code_generator_test.dart
+++ b/test/code_generator_test.dart
@@ -236,7 +236,7 @@ import 'dart:ffi' as ffi;
 
 /// Just a test struct
 /// heres another line
-class NoMember extends ffi.Struct{
+class NoMember extends ffi.Opaque{
 }
 
 class WithPrimitiveMember extends ffi.Struct{
@@ -388,6 +388,8 @@ typedef _dart_someFunc = ffi.Pointer<SomeStruc> Function(
       final struc_some = Struc(
         name: 'Some',
       );
+      final emptyGlobalStruc = Struc(name: 'EmptyStruct');
+
       final library = Library(
         name: 'Bindings',
         bindings: [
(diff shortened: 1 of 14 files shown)


test/code_generator_test.dart
```diff
@@ -436,7 +440,7 @@ final ffi.DynamicLibrary _dylib;
 /// The symbols are looked up in [dynamicLibrary].
 Bindings(ffi.DynamicLibrary dynamicLibrary): _dylib = dynamicLibrary;
 
-late final ffi.Pointer<int> _test1 = _dylib.lookup<ffi.Int32>('test1');
```
dcharkes, line 439: How did this not fail before?

dcharkes (approved):
lgtm

test/code_generator_test.dart
mannprerak2: In the last PR, Tim had updated this manually after generating the result and since the code_generator_tests compare the raw output by ffigen, `dartfmt` didn't run on this.

Perhaps I should update the code_generator_tests to completely generate bindings (including running dartfmt) and then compare the results? and should I do this in a new PR or this one itself?

dcharkes: Separate PR is fine.

Kinda nasty that dartfmt doesn't run on some things, otherwise we could just add tests that check that everything is formatted. Can you take a look at trying to run dartfmt on all Dart files in the repo on the CI and error out if one of them is not formatted as expected or fails to format. Maybe we can have a list of files that we should skip when checking so we can keep the CI green.

PR #143, opened 2021-01-25
Updated tests

mannprerak2:
- Updated code_generator_test to completely generate files instead of comparing the raw output.
- Throw `FormatException` if `dart format` is unable to format the generated bindings.


test/code_generator_tests/code_generator_test.dart
```diff
@@ -0,0 +1,382 @@
+// Copyright (c) 2020, the Dart project authors. Please see the AUTHORS file
```
dcharkes, line 1: 2021

dcharkes (approved):
Separate files is easier to read indeed. 👍

test/code_generator_tests/code_generator_test.dart
mannprerak2: Done.

PR #151, opened 2021-02-18
Bump version.

mannprerak2:
- Updated `package:quiver`'s version 
---
@dcharkes 
Looking at `package:quiver`'s [pubspec.yaml](https://github.com/google/quiver-dart/blob/a837ed13026cdeec95d4d935f3fe95fd55556183/pubspec.yaml#L9) I have updated the sdk constraints to `sdk: '>=2.12.0-0 <3.0.0'`.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -8,7 +8,7 @@ homepage: https://github.com/dart-lang/ffigen
 description: Generator for FFI bindings, using LibClang to parse C header files.
 
 environment:
-  sdk: '>=2.12.0-0 <3.0.0'
+  sdk: '>=2.12.0-259.9.beta <3.0.0'
 
 dependencies:
   ffi: ^1.0.0
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -3,18 +3,18 @@
 # BSD-style license that can be found in the LICENSE file.
 
 name: ffigen
-version: 2.0.0
+version: 2.0.1
 homepage: https://github.com/dart-lang/ffigen
 description: Generator for FFI bindings, using LibClang to parse C header files.
 
 environment:
-  sdk: '>=2.12.0-259.9.beta <3.0.0'
+  sdk: '>=2.12.0-0 <3.0.0'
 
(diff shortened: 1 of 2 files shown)


pubspec.yaml
```diff
@@ -3,18 +3,18 @@
...
 
 name: ffigen
-version: 2.0.0
+version: 2.0.1
 homepage: https://github.com/dart-lang/ffigen
 description: Generator for FFI bindings, using LibClang to parse C header files.
 
 environment:
-  sdk: '>=2.12.0-259.9.beta <3.0.0'
+  sdk: '>=2.12.0-0 <3.0.0'
```
dcharkes, line 11: Does this not include any releases for which the package or any of its dependencies don't work?

dcharkes (approved):

pubspec.yaml
dcharkes: Also, it looks like this SDK version on quiver as marked by pub as a dev release, while ours is marked as a beta release.

https://pub.dev/packages/quiver/versions
https://pub.dev/packages/ffigen/versions

mannprerak2: I actually saw the SDK constraints being specified like this in multiple other packages (args, yaml, test, charcode, etc) and thought this must help to automatically promote a preview release to a stable release once 2.12 stable sdk is released. (As mentioned at the end of this [medium article](https://medium.com/dartlang/preparing-the-dart-and-flutter-ecosystem-for-null-safety-e550ce72c010))

> Does this not include any releases for which the package or any of its dependencies don't work?

Do you mean we can't simply update `2.0.0` to `2.0.1` like this if we change the SDK constraints?

mannprerak2: Maybe I should revert this change for now.

mannprerak2: > Does this not include any releases for which the package or any of its dependencies don't work?

Ah, I get it now. Since struct by value is not supported by all releases, we can't update the constraints.
I have reverted the changes.

PR #148, opened 2021-02-12
Upgrade dependencies

mannprerak2:
Closes #128, Closes #109 
- Updated all dependencies to null safe versions.
- Cleanup: removed `dartfmt` usage since supported SDK versions are above 2.10.
- Update changelog and version (2.0.0, preview release)

PR #133, opened 2021-01-07
Support for Nested structs

mannprerak2:
Closes #4

- support for nested structs
- Updated readme, changelog, test
- Minor changes to remove null usage from code.

PR #154, opened 2021-02-23
Merge branch 'dart-dev' into 'master'.

### LINKED ISSUE #153: Merge dart-dev into master
With the Dart 2.12 beta soon moving to Dart 2.12 stable, we should merge our dart-dev branch (which is on beta) into the master branch.

https://github.com/dart-lang/ffigen/compare/dart-dev?expand=1

We can keep dart-dev around if we'd like to continue working on supporting `dart:ffi` features as they get available on dev. (Maybe we should delete the dart-master branch? If I recall correctly, running tests on master requires building the whole Dart SDK on the Ci just to run the tests.)

mannprerak2:
Closes #153


.github/workflows/test-package.yml
```diff
@@ -14,7 +14,6 @@ env:
 
 jobs:
   # Check code formatting and static analysis on a single OS (linux)
```
dcharkes, line 16: Send sentence with "." ?

dcharkes (approved):
lgtm

.github/workflows/test-package.yml
mannprerak2: Done. (I don't know why I always keep forgetting this 🙈)

dcharkes:
I prefer to not squash this one, because we already squashed the individual feature PRs on the dart-dev branch.

@mit-mit @athomas any objections?

mit-mit:
I don't have a lot of context here, so hard to comment. The preferred solution is a set of small targeted PRs, each of which can then be squashed.

dcharkes:
We've been extending this package based on new `dart:ffi` features being available on the Dart dev and beta releases on this branch. The master branch has been behind (no support for structs by value, nested structs, breaking API changes etc.)

We've made `package:ffi` pre-releases from this branch. If we squash and merge, we lose the source history for those releases. (I presume the pre-releases themselves are backed up on pub.dev itself.)

.github/workflows/test-package.yml
```diff
@@ -3,24 +3,23 @@ name: Dart CI
 on:
   # Run on PRs and pushes to the default branch.
   push:
-    branches: [ master ]
+    branches: [ dart-dev, master ]
```
mit-mit, line 6: Do we really want to run CI on this branch?

mit-mit:
Could we bring the changes over in a set of feature-based PRs?

.github/workflows/test-package.yml
mannprerak2: We have decided to keep the `dart-dev` branch for features that are only available on the dev SDK.

We can remove this, but I think it's better to keep it, or we'd have to change this line every time we pull changes between `maser` and `dart-dev` branch.

dcharkes: We want to have a solution for the common situation that:
1) We have a stable version of the package that uses the last Dart stable release and someone makes a PR for that.
2) We have a dev version of the package that uses the last Dart dev release (including new dart:ffi features) and someone makes a PR for that.

So, I think we need two branches.

dcharkes:
We could, but this would be a recurring issue every stable release.

1. We have a stable version of the package that uses the last Dart stable release and someone makes a PR for that.
2. We have a dev version of the package that uses the last Dart dev release (including new dart:ffi features) and someone makes a PR for that.

And then every stable release, we would have to redo all the PRs from the dev branch to the stable branch. Wouldn't it be cleaner to have just a merge from dev to stable every Dart stable release?

mannprerak2:
@dcharkes  Should I squash the last 2 commits into 1 commit?

dcharkes:
Alternative approach as suggested by @mit is to have the master branch track dart dev releases and do all the PRs there. And then when we make a stable release make the master branch track dart stable (or beta).

As for getting this work on to the master branch, we could make individual PRs grouping commits on this branch.

* Support for Nested structs (#133) …
* Support for passing/returning Structs by value. (#134) …
* Cleanups PR
  * Remove setup phase (#135)
  * Removed the usage of --no-sound-null-safety (#137) …
* Globals (#139)
* Use Opaque to represent empty Structs (#142) …
   * (should be part of whatever PR the tests should have been updated in) Updated tests (#143) …
* Skip inline functions, added test, updated changelog and version. (#147)
* Stable release PR
  * Upgrade dependencies (#148)
  * Out of experimental …
  * Bump version. (#151) …
  * Merge branch 'master' of github.com:dart-lang/ffigen into dart-dev

dcharkes:
After discussing with @natebosch, we came up with this: https://github.com/dart-lang/ffigen/wiki/Branches.

dcharkes:
Done in separate PRs.

PR #163, opened 2021-03-01
Support for passing/returning Structs by value. (#134)

mannprerak2:
* Added support for structs by value, fixed tests, and example
* updated version, readme, changelog
* Added more tests
* Added struct return by value to native test
* Fix struct by value in nested typedefs


dcharkes (approved):
Reviewed on https://github.com/dart-lang/ffigen/pull/134.

PR #166, opened 2021-03-01
Use Opaque to represent empty Structs, update tests (#142, #143)

mannprerak2:


example/c_json/pubspec.yaml
```diff
@@ -5,7 +5,7 @@
 name: c_json_example
 
 environment:
-  sdk: '>=2.12.0-198.0.dev <3.0.0'
+  sdk: '>=2.12.0-230.0.dev <3.0.0'
```
dcharkes, line 8: Should this also be 237? (Also in other places.)

dcharkes (approved):
Already reviewed in #142 and #143.

example/c_json/pubspec.yaml
mannprerak2: Done.

mannprerak2: This hint for empty structs was added in 230 so I don't think this should of any issue.

PR #174, opened 2021-03-04
Support for Dynamic linking

### LINKED ISSUE #127: Null-safety: Switch back to stable branch.
We currently test on the beta branch. This should be reverted when null safety becomes stable.

### LINKED ISSUE #173: Support dynamic linking
We should support dynamic linking and accept just the `lookup` function of a `DynamicLibrary` in the constructor.

```dart
  final ffi.Pointer<T> Function<T extends ffi.NativeType>(String symbolName)
      _lookup;

  BoringSsl(ffi.DynamicLibrary dynamicLibrary)
      : _lookup = dynamicLibrary.lookup;

  /// The symbols are looked up with [lookup].
  BoringSsl.fromLookup(
      ffi.Pointer<T> Function<T extends ffi.NativeType>(String symbolName)
          lookup)
      : _lookup = lookup;
```

The bindings should then be changed to use `_lookup<NativeFunction<...>>('...').asFunction<...>()` instead of `_dylib.lookupFunction<...,...>('...')`.

By making these changes by hand, I was able to use `package:ffigen` for `package:webcrypto`. Webcrypto uses [dynamic linking](https://github.com/google/webcrypto.dart/blob/master/lib/src/boringssl/lookup/lookup_symbol_flutter.dart#L29-L41).

cc @jonasfj

@mannprerak2, I could make a PR maybe later this week. But if you want to go ahead and do it, that is fine too.

mannprerak2:
Closes #173, Closes #127
- Added lint rule `unnecessary_brace_in_string_interps`.
- Added `.fromLookup` constructor to support dynamic linking. (173)
- Updated version (`2.0.3 => 2.1.0`, changelog, examples, test.
- Updated travis to target version`2.12.0` (127).
- Updated SDK constraints to `>=2.12.0 <3.0.0`.

---

I am not sure why we were suddenly getting `info - Avoid using braces in interpolation when not needed at ...` [on Travis after pushing to master](https://github.com/dart-lang/ffigen/runs/2023575453), but I went ahead and added this lint rule.

### DIFF
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        sdk: [2.12.0-259.9.beta] # TODO(127): Revert to stable.
+        sdk: [2.12.0]
     steps:
       - uses: actions/checkout@v2
       - uses: dart-lang/setup-dart@v1.0
@@ -43,7 +43,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: dart-lang/setup-dart@v1.0
         with:
-          sdk: 2.12.0-259.9.beta # TODO(127): Revert to stable.
+          sdk: 2.12.0
       - name: Install dependencies
         run: dart pub get
       - name: Install libclang-10-dev
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 2.0.4
+- Support for dynamic linking, use `NativeLibrary.fromLookup()` constructor.
+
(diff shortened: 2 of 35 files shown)


dcharkes (approved):
> I am not sure why we were suddenly getting `info - Avoid using braces in interpolation when not needed at ...` [on Travis after pushing to master](https://github.com/dart-lang/ffigen/runs/2023575453), but I went ahead and added this lint rule.

Me neither, but thanks for cleaning it up! 💯 

> * Added `.fromLookup` constructor to support dynamic linking. (173)

🚀  🚀  🚀

PR #181, opened 2021-03-05
Added key `structs -> dependency-only` to generate Opaque struct dependencies.

### LINKED ISSUE #176: Only generate non-included structs as Opaque if only referenced by pointer
https://github.com/dart-lang/ffigen/blob/89084d968ce0d587b19290366762ab7af7ffbe61/README.md#L474-L477

If `struct`s are only referenced by `Pointer<struct>` we should generate them as `Opaque` if they are not included.

For example [here](https://github.com/google/webcrypto.dart/blob/b5c79268214ac1af2b1835dbc3a0584cdb436d79/lib/src/third_party/boringssl/generated_bindings.dart#L3646-L3683).

If someone wants the full struct, they can include it.

mannprerak2:
Closes #176 
- Added config key `struct-dependencies` (Options - full (default) | opaque) to generate empty `Opaque` structs when structs were excluded and only passed by reference.
- Update readme, version, changelog, faq.
- Added test.

---

@dcharkes 
I've currently added top-level key `struct-dependencies` (https://github.com/dart-lang/ffigen/issues/176#issuecomment-791572961)
But if you think nesting might be better let me know I'll update the PR to reflect that.


dcharkes:
> I've currently added top-level key `struct-dependencies` ([#176 (comment)](https://github.com/dart-lang/ffigen/issues/176#issuecomment-791572961))

It's a property about structs right, so maybe nesting is cleaner to not get too many top level entries in the config. Of course it is talking about the structs as a dependency, not the dependencies of structs, which it can be through a function type rather than another struct.

Maybe we should make it `dependency-only` or `struct-depedency-only` to signify that this setting is only applicable to structs which are only a dependency and not included by hand.

```yaml
structs:
  dependency-only: full | opaque
```

wdyt?

lib/src/header_parser/sub_parsers/structdecl_parser.dart
```diff
@@ -83,15 +91,30 @@ Struc? parseStructDeclaration(
...
-    _stack.top.struc = bindingsIndex.getSeenStruct(structUsr)!;
+    _stack.top.struc = bindingsIndex.getSeenStruct(structUsr);
+
+    final skipDependencies = _stack.top.struc!.parsedDependencies ||
+        (config.structDependencies == StructDependencies.opaque &&
+            pointerReference &&
+            ignoreFilter);
+
+    if (!skipDependencies) {
+      // Prevents infinite recursion if struct is member of itself.
```
dcharkes, line 105: It can't be a member of itself (nesting). But a struct type can have a pointer member pointing to its own type.

dcharkes (approved):
lgtm

lib/src/header_parser/sub_parsers/structdecl_parser.dart
mannprerak2: Yeah, definitely that's what I meant here. I should update the comment.

mannprerak2:
```yaml
structs:
  dependency-only: full | opaque
```

Done.

PR #187, opened 2021-03-08
Fix `sort: true` not working, remove //+ from full comments.

### LINKED ISSUE #185: sort: true not working?
> Sort the bindings according to name.

I don't see the function bindings (or the structs) being sorted.

Both function bindings:

```dart
  /// // BN_new creates a new, allocated BIGNUM and initialises it.
  ffi.Pointer<BIGNUM> BN_new() {
    return _BN_new();
  }

  late final _BN_new_ptr = _lookup<ffi.NativeFunction<_c_BN_new>>('BN_new');
  late final _dart_BN_new _BN_new = _BN_new_ptr.asFunction<_dart_BN_new>();

  /// // BN_free frees the data referenced by |bn| and, if |bn| was originally
  /// // allocated on the heap, frees |bn| also.
  void BN_free(
    ffi.Pointer<BIGNUM> bn,
  ) {
    return _BN_free(
      bn,
    );
  }

  late final _BN_free_ptr = _lookup<ffi.NativeFunction<_c_BN_free>>('BN_free');
  late final _dart_BN_free _BN_free = _BN_free_ptr.asFunction<_dart_BN_free>();
```

And structs:

```dart
class FILE extends ffi.Opaque {}

class EVP_CIPHER extends ffi.Struct {
  // ...
}
```

https://github.com/google/webcrypto.dart/blob/use-ffigen-2/lib/src/third_party/boringssl/generated_bindings.dart

### LINKED ISSUE #186: /// // comments
```dart
  /// // BN_free frees the data referenced by |bn| and, if |bn| was originally
  /// // allocated on the heap, frees |bn| also.
  void BN_free(
    ffi.Pointer<BIGNUM> bn,
  ) {
    return _BN_free(
      bn,
    );
  }
```

The comments copying includes the `//`'s which were already in the C source. Would it be possible to detect and remove these?

mannprerak2:
Closes #185, Closes #186
- Fix `sort: true` not working.
- Remove 2 or more forward slashes `/` from full comments.
- Added test.
- Update changelog and version.


mannprerak2:
Should've added a test before, this went un-noticed since acef4b48b06caecc531d105ee01c2fc9f9efe784 (8 mos).

mannprerak2:
@dcharkes generated typedefs may not always be in a sorted order (e.g typedefs of dependencies are also added at the same time) should we also sort those?

lib/src/header_parser/utils.dart
```diff
@@ -197,8 +197,9 @@ String? removeRawCommentMarkups(String? string) {
   string = string.replaceAll('/*', '');
   string = string.replaceAll('*/', '');
 
-  // Remove any *'s in the beginning of a every line.
+  // Remove any *'s or // in the beginning of a every line.
```
jonasfj, line 200: nit: it might someday be worth doing a bit of a heuristic here.

Like:
 * If comment starts with `\s*/\**?` then:
    * remove `\s*/\**?` from the start of the comment,
    * remove `*/\s*` at the end of the comment,
    * if every line starts with `\s\*` then remove `\s\*` from start of every line.
 * If comment starts with `\s*///?` then remove `\s*///?` from start of every line.

dcharkes, line 200: This didn't cause any of our existing large tests to get different output? Did you regenerate all of those?

If not, maybe we should add some more small manual tests with different comment styles.

dcharkes (commented):
Thanks for the quick PR! 😃 

We should have some test coverage for the different comment styles.

lib/src/header_parser/utils.dart
mannprerak2: The large tests use ** style so it didn't cause any different output. We are testing the comments under large tests. LibClang tests - doxygen, full and SQLIte tests any, full comments.

mannprerak2: I'll add new tests and update the way we remove comment markup using suggestions from Jonas.

mannprerak2: Done.

mannprerak2: Done.

test/header_parser_tests/comment_markup.h
```diff
@@ -0,0 +1,26 @@
...
+ */
+void com3();
+
+// Test comment for struct.
+struct com4{
+    /// Muli-line test comment for struct field
+    // With multiple line and both // and ///.
+    int a;
+
+    /* Single line field comment.*/
```
dcharkes, line 24: `/* Single line field comment.*/` -> `/* Single line field comment. */`

dcharkes (approved):
lgtm!

test/header_parser_tests/comment_markup.h
mannprerak2: Done.

Issue #190, opened 2021-03-09
automatically add C standard library headers to compiler-opts


dcharkes:
On MacOS, the std lib is not on the include path by default.

We should consider adding logic for searching for it in this package, and adding it to the compiler options by default.

Every project is going to have to add these to the compiler opts otherwise.

For example: https://github.com/dart-lang/ffigen/blob/master/test/large_integration_tests/large_test.dart#L82-L83

mannprerak2:
These headers seem to be available at 2 locations -
1. `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/12.0.0/include/`
2. If command line tools are installed `/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/Kernel.framework/Headers/`

We could have used **`2`** but it may not be present if the user hasn't installed command-line tools, so we can possibly only use `1` as the default location but would need to fix the **`12.0.0`** to whatever is available in the path.

dcharkes:
We could search for versions instead of pinning them

```dart
    const brewLlvmPath = "/usr/local/opt/llvm/lib/clang";
    const commandLineToolsClangPath =
        "/Library/Developer/CommandLineTools/usr/lib/clang";
    const searchPaths = [brewLlvmPath, commandLineToolsClangPath];
    List<String> includePaths = [];
    for (final searchPath in searchPaths) {
      final result = Process.runSync('ls', [searchPath]);
      final stdout = result.stdout as String;
      if (stdout != "") {
        final versions = stdout.split("\n").where((s) => s != "");
        for (final version in versions) {
          final path = "$searchPath/$version/include";
          includePaths.add(path);
        }
      }
    }
    if (includePaths.isEmpty) {
      print("The stdlib headers could not be found.");
      print("Paths searched: $searchPaths");
      print("Install Command Line Tools.");
      print("Or, please run `brew install llvm`.");
    }
```

dcharkes:
What about this API?

```yaml
automatic-compiler-opts:
  macos:
    # Looks for the c standard library headers in CommandLineTools, XCode, and llvm.
    # Issues a warning if headers cannot be found.
    # Automatically adds `-Ipath/to/c-standard-library` to `compiler-opts`.
    include-c-standard-library: true (default) | false
```

Sunbreak:
Is https://github.com/dart-lang/ffigen/issues/170 related?

[...]

PR #192, opened 2021-03-10
Added new command-line option `--compiler-opts`.

### LINKED ISSUE #189: pass in compiler-opts on commandline
Depending on host OS and installed toolchains we need to pass in different compiler options to make various ffigen invocations work.

Instead of adding all individual versions to the `yaml` configuration, it might make sense to allow passing these options in to the ffigen invocation.

Instead of

```yaml
compiler-opts: '-I/usr/local/opt/llvm/lib/clang/11.1.0/include/' # 11.1.0 might be some other version
```

write a script

```dart
import 'dart:io';

void main(List<String> arguments) {
  final headers = headerPaths();
  final root =
      Directory.fromUri(Platform.script).parent.parent.uri.toFilePath();
  final result = Process.runSync('dart', [
    'run',
    'ffigen',
    '--config=$root/bin/ffigen.yaml',
    ...headers, // This is currently not supporteed
  ]);
}

List<String> headerPaths() {
  if (Platform.isMacOS) {
    const brewLlvmPath = "/usr/local/opt/llvm/lib/clang";
    final result = Process.runSync('ls', [brewLlvmPath]);
    final stdout = result.stdout as String;
    if (stdout == "") {
      print("The stdlib headers could not be found.");
      print("Please run `brew install llvm`.");
      exit(1);
    }

    // We pick the first available one.
    final version = stdout.split("\n").first;

    final path = "$brewLlvmPath/$version/include";
    final include = "-I$path";
    return [include];
  }
  throw UnsupportedError("Unknown operating system");
}
```

(Without providing headers `stdarg.h` and other headers from the standard library cannot be found by clang on various project.)

Maybe we should interpret all arguments that we don't consume as `compiler-opts`.

mannprerak2:
Closes #189 
- Added new command-line option --compiler-opts.
- Added test, update changelog, version, readme.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 # 2.2.5
-- Added new command line flag `--compiler-opts` to ffigen.
+- Added new command line flag `--compiler-opts` to the command line tool.
 
 # 2.2.4
 - Fix `sort: true` not working.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 2.2.5
+- Added new command line flag `--compiler-opts` to ffigen.
+
 # 2.2.4
 - Fix `sort: true` not working.
 - Fix extra `//` or `///` in comments when using `comments -> style`: `full`.
--- a/README.md
+++ b/README.md
@@ -135,11 +135,16 @@ description: 'Bindings to SQLite'
   </tr>
   <tr>
     <td>compiler-opts</td>
-    <td>Pass compiler options to clang.</td>
+    <td>Pass compiler options to clang. You can also pass
+    these via the command line tool.</td>
     <td>
 
 ```yaml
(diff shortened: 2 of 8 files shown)


mannprerak2:
@dcharkes should I also resolve #190 here?

dcharkes (approved):
lgtm

dcharkes:
> @dcharkes should I also resolve #190 here?

Whatever you prefer, a separate PR is fine for me as well.

mannprerak2:
Let's do a separate PR then, I hadn't done any work on that till now.

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 2.2.5
+- Added new command line flag `--compiler-opts` to ffigen.
```
dcharkes, line 2: to ffigen. -> to the command line tool.

mannprerak2: Done

PR #193, opened 2021-03-12
Automatically add std lib for macos, allow passing list to compiler-opts.

### LINKED ISSUE #190: automatically add C standard library headers to compiler-opts

mannprerak2:
Closes #190 
- Added config key `compiler-opts-automatic -> macos -> include-c-standard-library`.
- Allow passing list to compiler-opts.
- Update test, version, changelog, readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ description: 'Bindings to SQLite'
     <td>
 
 ```yaml
-compiler-opts: # You can also pass in a string.
+compiler-opts:
   - '-I/usr/lib/llvm-9/include/'
 ```
 and/or via the command line -
--- a/README.md
+++ b/README.md
@@ -140,12 +140,28 @@ description: 'Bindings to SQLite'
     <td>
 
 ```yaml
-compiler-opts: '-I/usr/lib/llvm-9/include/'
+compiler-opts: # You can also pass in a string.
+  - '-I/usr/lib/llvm-9/include/'
 ```
 and/or via the command line -
 ```bash
 dart run ffigen --compiler-opts "-I/headers
 -L 'path/to/folder name/file'"
+```
+  </td>
+  </tr>
+    <tr>
+    <td>compiler-opts-automatic -> macos -> include-c-standard-library</td>
+    <td>Tries to automatically find and add C standard library path to
+    compiler-opts on macos.
+    <b>Default: true</b>
+    </td>
+    <td>
+
+```yaml
+compiler-opts-automatic:
+  macos:
+    include-c-standard-library: false
 ```
   </td>
   </tr>
--- a/lib/src/config_provider/path_finder.dart
+++ b/lib/src/config_provider/path_finder.dart
@@ -1,0 +1,51 @@
+// Copyright (c) 2021, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:logging/logging.dart';
+
+/// Utils for finding headers on system.
+final _logger = Logger('ffigen.config_provider.path_finder');
+
+List<String> getHeadersForMac() {
+  final includePaths = <String>[];
+
+  /// If CommandLineTools are installed use those headers.
+  const cmdLineToolHeaders =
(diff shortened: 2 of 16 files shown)


test/large_integration_tests/large_test.dart
```diff
@@ -79,8 +79,6 @@ ${strings.name}: SQLite
 ${strings.description}: Bindings to SQLite.
 ${strings.output}: unused
 ${strings.arrayWorkaround}: true
-# Needed for stdarg.h in MacOS
```
dcharkes, line 82: Nice to see these cleanups! 👍

README.md
```diff
@@ -140,12 +140,28 @@ description: 'Bindings to SQLite'
     <td>
 
 ```yaml
-compiler-opts: '-I/usr/lib/llvm-9/include/'
+compiler-opts: # You can also pass in a string.
```
dcharkes, line 143: `# You can also pass in a string.`

Maybe just remove this.

dcharkes (approved):
Sweet! 🍨

README.md
mannprerak2: Done.

I noticed I had set my CPATH variable but many users wouldn't have, so I've added this location `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include`
in the include path as well.

These are the system headers that are only available in this location.

lib/src/config_provider/path_finder.dart
```diff
@@ -0,0 +1,63 @@
...
+import 'dart:io';
+
+import 'package:logging/logging.dart';
+import 'package:path/path.dart' as p;
+
+final _logger = Logger('ffigen.config_provider.path_finder');
+
+/// This will return include path from either LLVM, XCode or CommandLineTools.
+List<String> getHeadersForMac() {
+  final includePath = <String>[];
```
dcharkes, line 16: includePaths ?

lib/src/config_provider/path_finder.dart
```diff
@@ -0,0 +1,63 @@
...
+
+  /// Add system headers.
+  const systemHeaders =
+      '/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include';
+  if (Directory(systemHeaders).existsSync()) {
+    _logger.fine('Added $systemHeaders to compiler-opts.');
+    includePath.add('-I' + systemHeaders);
```
dcharkes, line 23: These system headers partially overlap with the C standard library (e.g. no stdarg.h, but yes stddef.h), is there any risk in including both?

Maybe system headers should be under a different key in the config yaml and be not included by default?

lib/src/config_provider/path_finder.dart
```diff
@@ -0,0 +1,63 @@
...
+
+import 'dart:io';
+
+import 'package:logging/logging.dart';
+import 'package:path/path.dart' as p;
+
+final _logger = Logger('ffigen.config_provider.path_finder');
+
+/// This will return include path from either LLVM, XCode or CommandLineTools.
+List<String> getHeadersForMac() {
```
dcharkes, line 15: getCStandardLibraryHeadersForMac() maybe to exclude any other type of headers?

dcharkes (approved):
Overall lgtm, some questions about compatibility though.

lib/src/config_provider/path_finder.dart
mannprerak2: > is there any risk in including both?

I am actually seeing some clash between CommandLineTools and System headers. If I add CommandLineTools to include path first and then add System headers, then I am seeing errors like `int32_t` not defined, `size_t` not defined.
My guess is that CommandLine Tools doesn't define these and set the include guard macro?

So we definitely need to add the system headers to path first incase we added commandlinetools.

mannprerak2: Also, CPATH is looked at after the compiler opt paths, so the user would need to specify system headers in compiler opts in case the C standard library we used was from CommandLineTools.

dcharkes: Thanks for investigating!

This PR works on my side. 👍

PR #194, opened 2021-03-15
Added key `llvm-path`. Deprecate `llvm-lib`.

### LINKED ISSUE #156: Inconsistency and restricted usage of `llvm-lib` config.
The `llvm-lib` config was meant for users to specify `llvm/lib` folder in case they didn't install `llvm` in its default location. This location is then used to access the dynamic library required for `ffigen` to run.

**Issues** -
- This is inconsistent since `libclang.dll` exists in`llvm\bin` folder on windows (for some mystical reason).
- `llvm-lib` accepts a single `string`, but its possible that multiple users may be working on the same project and have llvm installed at different locations. This will lead to conflict and they'd need to modify this value to be able to run ffigen.

**Solution** -
We can update `llvm-lib` to `llvm-path` so they only need to specify `path/to/llvm`. We can allow `llvm-path` to be an array of strings and try to find the dynamic library in one of those paths specified.

This is a breaking change, but a very small one, and since most users don't need this config we can probably make do with incrementing our version from `2.X.0` to `2.Y.0`.

what do you think @dcharkes ?

mannprerak2:
Close #156
- Added key `llvm-path`. This takes a list of `path/to/llvm`.
- Deprecated key `llvm-lib`
- Update version, changelog, readme.

### DIFF
--- a/lib/src/config_provider/config.dart
+++ b/lib/src/config_provider/config.dart
@@ -181,10 +181,20 @@ class Config {
   /// Key: Name, Value: [Specification]
   Map<List<String>, Specification> _getSpecs() {
     return <List<String>, Specification>{
+      //TODO: Deprecated, remove in next major update.
       [strings.llvmLib]: Specification<String>(
         requirement: Requirement.no,
         validator: llvmLibValidator,
         extractor: llvmLibExtractor,
+        defaultValue: () => '',
+        extractedResult: (dynamic result) {
+          if ((result as String).isNotEmpty) _libclangDylib = result;
+        },
+      ),
+      [strings.llvmPath]: Specification<String>(
+        requirement: Requirement.no,
+        validator: llvmPathValidator,
+        extractor: llvmPathExtractor,
         defaultValue: () => findDylibAtDefaultLocations(),
         extractedResult: (dynamic result) => _libclangDylib = result as String,
       ),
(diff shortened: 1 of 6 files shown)


lib/src/strings.dart
```diff
@@ -23,6 +23,10 @@ String get dylibFileName {
 }
 
 const llvmLib = 'llvm-lib';
+const llvmPath = 'llvm-path';
+
+/// Name of the parent folder of dynamic library `lib` or `bin`(on windows).
```
dcharkes, line 28: space between `bin` and (on

lib/src/config_provider/config.dart
```diff
@@ -181,12 +181,25 @@ class Config {
   /// Key: Name, Value: [Specification]
   Map<List<String>, Specification> _getSpecs() {
     return <List<String>, Specification>{
+      //TODO: Deprecated, remove in next major update.
```
dcharkes, line 184: Emit a deprecation warning when used.

dcharkes (approved):

lib/src/config_provider/spec_utils.dart
```diff
@@ -351,6 +351,8 @@ String llvmLibExtractor(dynamic value) {
 }
 
 bool llvmLibValidator(List<String> name, dynamic value) {
+  _logger.severe(
```
mannprerak2, line 354: @dcharkes We are emitting a warning here.

dcharkes: My bad. Maybe just make it a warning instead of severe.

mannprerak2: Done.

PR #200, opened 2021-04-05
Fix error caused by duplicate declaration names and collision with ffi library prefix

### LINKED ISSUE #199: Duplicate class and function name
I was trying to use Wireguard from Dart using this https://git.zx2c4.com/wireguard-tools/tree/contrib/embeddable-wg-library

But the generated output gave both a function and a class with the same name.
There was the function:
```
  ffi.Pointer<if_nameindex> if_nameindex() {
    return _if_nameindex();
  }
```
and the class
```
class if_nameindex extends ffi.Struct {
  @ffi.Uint32()
  external int if_index;

  external ffi.Pointer<ffi.Int8> if_name;
}
```

Which leads to the error that "if_nameindex isn't a type".
To fix it I had to manually rename the class to `class IfNameIndex` and it's references.

Should this be automatically handled by `ffigen`?
I saw the rename option in the docs, but couldn't figure out if that would work, maybe some better doc there?

mannprerak2:
Closes #199 

- Fix error caused by duplicate declaration names across different namespaces (global and inside NativeLibrary).
- Fix name conflict of lookup bindings with ffi library prefix.
- Updated tests, version, and changelog.

---

Fixes for some corner cases of name conflicts that we missed earlier.

### DIFF
--- a/test/collision_tests/decl_decl_collision_test.dart
+++ b/test/collision_tests/decl_decl_collision_test.dart
@@ -44,6 +44,12 @@ void main() {
             returnType: Type.nativeType(SupportedNativeType.Void)),
         Constant(name: 'testCrossDecl', rawValue: '0', rawType: 'int'),
         EnumClass(name: 'testCrossDecl'),
+
+        /// Conflicts with ffi library prefix, name of prefix is changed.
+        Struc(name: 'ffi'),
+        Func(
+            name: 'ffi_1',
+            returnType: Type.nativeType(SupportedNativeType.Void)),
       ]);
       final l2 = Library(name: 'Bindings', bindings: [
         Struc(name: 'TestStruc'),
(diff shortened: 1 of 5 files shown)


mannprerak2:
Marking this as a draft so we merge #201 first.

test/collision_tests/decl_decl_collision_test.dart
```diff
@@ -36,6 +36,20 @@ void main() {
           rawType: 'int',
           rawValue: '0',
         ),
+
+        /// Conflicts across declarations
```
dcharkes, line 40: .

dcharkes (approved):

test/collision_tests/decl_decl_collision_test.dart
mannprerak2: Fixed 😄 .

dcharkes:
Good stuff! 🚀

PR #205, opened 2021-04-16
Support for Packed struct

### LINKED ISSUE #204: packed structs
`2.13.0-211.6.beta` has support for packed structs (https://github.com/dart-lang/sdk/issues/38158).

Can we detect with clang a struct is packed? and add the right annotation to it?

mannprerak2:
Closes #204 

- Added support for `@Packed(X)` Struct annotation.
- Added config key `structs -> pack` to override the pack values.
- Updated SDK constraints to `>=2.13.0-211.6.beta`.
- Updated version to `3.0.0-beta.1`
- Added tests, Updated changelog.


mannprerak2:
@dcharkes 
From our plan https://github.com/dart-lang/ffigen/issues/204#issuecomment-820390852, It seems like we cannot reliably get a align value if only `__attribute__((aligned(X))` is used. Since it could be inside a `#pragma` which changes its behavior to be the exact opposite.

I have added some tests and it all seems to be working. I'll try to do some manual testing on Win32 as you suggested.

There's a config key `struct -> pack` that can override the generated value. It targets the generated name instead of the original name (in the C header). I think this makes more sense since we are overriding the value which would be easier done by the user by directly matching with the generated name, this has also been documented in the readme.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# 3.0.0-beta.1
```
dcharkes, line 1: Is it customary the bump the major revision when we update the SDK constraint?

The pub solver will opt out people automatically from a 2.5.0 if they are not on 2.13.

And we're not doing an breaking changes I believe.

lib/src/config_provider/config_types.dart
```diff
@@ -6,7 +6,6 @@
 import 'dart:io';
 
 import 'package:quiver/pattern.dart' as quiver;
-
```
dcharkes, line 9: by accident?

lib/src/header_parser/sub_parsers/structdecl_parser.dart
```diff
@@ -33,6 +34,39 @@ class _ParsedStruc {
       (dartHandleMember && config.useDartHandle) ||
       incompleteStructMember;
 
+  // A struct without any attribute is definitely not packed. #pragma pack(...)
+  // also adds an attribute, but it's unexposed and cannot be travesed.
```
dcharkes, line 38: Please document in the readme which of the two ways clang allows to detect in which ones we don't. Then users have a starting point for manually verifying whether it did the right thing for them.

lib/src/header_parser/sub_parsers/structdecl_parser.dart
```diff
@@ -33,6 +34,39 @@ class _ParsedStruc {
...
+  bool hasAttr = false;
+  // A struct which as a __packed__ attribute is definitely packed.
+  bool hasPackedAttr = false;
+  // Stores the max align value from all the children.
+  int maxChildAlignValue = 0;
+  // Align value of this struct.
```
dcharkes, line 44: Alignment of this struct.

lib/src/header_parser/sub_parsers/structdecl_parser.dart
```diff
@@ -33,6 +34,39 @@ class _ParsedStruc {
       (dartHandleMember && config.useDartHandle) ||
       incompleteStructMember;
 
+  // A struct without any attribute is definitely not packed. #pragma pack(...)
+  // also adds an attribute, but it's unexposed and cannot be travesed.
+  bool hasAttr = false;
+  // A struct which as a __packed__ attribute is definitely packed.
+  bool hasPackedAttr = false;
+  // Stores the max align value from all the children.
```
dcharkes, line 42: max align value -> maximum alignment

lib/src/header_parser/utils.dart
```diff
@@ -236,6 +236,10 @@ extension CXTypeExt on clang_types.CXType {
     return clang.clang_getTypeKindSpelling(kind()).toStringAndDispose();
   }
 
+  int align() {
```
dcharkes, line 239: `alignment()` in Dart we don't really abbreviate method names. And this looks up the alignment rather than aligning something.

You could even do `int get alignement =>` if you want to be fancy. 😄

test/header_parser_tests/expected_bindings/_expected_packed_structs_bindings.dart
```diff
@@ -0,0 +1,52 @@
...
+}
+
+/// Should be packed with 2.
+@ffi.Packed(2)
+class Pack2WithPragma extends ffi.Struct {
+  @ffi.Int32()
+  external int a;
+}
+
+/// Should be packed with 8.
```
dcharkes, line 39: ?

tool/libclang_config.yaml
```diff
@@ -76,6 +76,7 @@ functions:
     - clang_getPointeeType
     - clang_getCanonicalType
     - clang_Type_getNamedType
+    - clang_Type_getAlignOf
```
dcharkes, line 79: Every time you use a new part of the clang API by using ffigen itself it makes me happy! 😄

dcharkes (approved):

tool/libclang_config.yaml
mannprerak2: Haha, me too 😆

CHANGELOG.md
mannprerak2: Right.

lib/src/header_parser/utils.dart
mannprerak2: Updated `align()` to -> `alignment()`

> You could even do int get alignment...

Yeah, I thought that too, but then it wouldn't be consistent with the code earlier.

dcharkes: If you'd like, you can always send a refactoring PR to make it more pretty. I don't have a very strong opinion about this. :)

test/header_parser_tests/expected_bindings/_expected_packed_structs_bindings.dart
mannprerak2: Turns out that #pragma pack(X) sets the Upper bound value of the alignment. (While the alignment(X) attribute sets the lower bound.) I have updated the test accordingly.

lib/src/header_parser/utils.dart
mannprerak2: Yeah, I might do that someday. There's a lot I think which can be improved.

lib/src/header_parser/sub_parsers/structdecl_parser.dart
mannprerak2: Updated reamde. => https://github.com/mannprerak2/ffigen/tree/packed_struct#user-content-configurations:~:text=Note%3A%20Ffigen%20can%20only%20reliably%20identify,in%20which%20case%20you%20can%20override

mannprerak2:
I ran this on `windows.h` (Win32) and the Pragma values seem to match those in https://github.com/timsneath/win32 🎉

dcharkes:
Awesome! Thanks @mannprerak2 !

PR #206, opened 2021-04-16
Support for Inline Array in Struct

### LINKED ISSUE #5: Support for inline arrays in structs
Waiting on support for [inline arrays in Structs](https://github.com/dart-lang/sdk/issues/35763) in `dart:ffi`.
Currently, we generate Array helpers as a workaround (Use `array-workaround: true` to enable), this will be replaced when proper inline
array support arrives.

mannprerak2:
Closes #5 

- Added support for generating Inline Arrays in Struct.
- Removed `array-workaround` config key. (Using this will emit a warning -> `unknown key in config`).
- Updated test, changelog, readme, and version.


example/libclang-example/generated_bindings.dart
```diff
@@ -7928,64 +7928,8 @@ abstract class CXGlobalOptFlags {
...
-        break;
-      case 2:
-        _struct._unique_data_item_2 = value;
-        break;
-      default:
-        throw Exception('Invalid Array Helper generated.');
-    }
-  }
+  @ffi.Array.multi([3])
+  external ffi.Array<ffi.Uint64> data;
```
dcharkes, line 7932: Since this also provides `[]` and `[]=` this is actually not an API change at all, that is pretty sweet!

test/native_test/native_test.dart
```diff
@@ -91,19 +91,20 @@ void main() {
     test('double', () {
       expect(bindings.Function1Double(0), 42.0);
     });
-    test('array-workaround: Order of access', () {
+    test('Array Test: Order of access', () {
       final struct1 = bindings.getStruct1();
       var expectedValue = 1;
-      for (var i = 0; i < struct1.ref.data.dimensions[0]; i++) {
```
dcharkes, line 97: Ah the old api gave access to `.dimensions`? Then this is kind of a breaking change.

But we should likely add that kind of access to `Array` in `dart:ffi` as well.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 2.5.0-beta.2
+- Added support for Inline Arrays in Struct.
```
dcharkes, line 2: Inline Arrays -> inline arrays  (It's not two names, so they don't need to be capitalized).
Struct -> structs -> `Struct`s.

dcharkes (approved):
Sweet! I love how many lines of code we delete here (both generated and non generated)!

test/native_test/native_test.dart
mannprerak2: >  Then this is kind of a breaking change

Should I update the version to 3.0.0-beta then?

> But we should likely add that kind of access to Array in dart:ffi as well.

Yeah, I was wondering that there's no way to get the length of an array. (even though its constant and the user would know it beforehand)

dcharkes: Hm, if we want to be true to semantic versioning, we probably should make it `3.0.0-beta.0`, good catch!

dcharkes: Also, the key is probably rejected in the config right now, instead of just deprecated. So that is also a breaking change.

mannprerak2: > the key is probably rejected

It will simply warn about an unknown key called `array-workaound`, the tool would still work since an unknown key is not a severe error.

---

Updated version.

mannprerak2: I've also removed the previously deprecated `llvm-lib` config.

PR #209, opened 2021-04-28
Update readme to clarify C support.

### LINKED ISSUE #208: Clarify C++ support
I tried to run ffigen on some mixed c/c++ codebase to extract the c-bindings.

I got very wild errors and apparently ffigen just doesn’t support c++ headers (yet). Can you please clarify this?

mannprerak2:
Closes #208

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 Binding generator for [FFI](https://dart.dev/guides/libraries/c-interop) bindings.
 
-> Note: ffigen and `dart:ffi` only support `C` at the moment.
+> Note: ffigen only supports parsing `C` headers at the moment.
 
 ## Example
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,6 +4,8 @@
 
 Binding generator for [FFI](https://dart.dev/guides/libraries/c-interop) bindings.
 
+> Note: ffigen only supports parsing `C` headers at the moment.
```
dcharkes, line 7: Leave out the "at the moment".

We will never support C++ headers, because C++ does not have a stable ABI like C (or use of `extern "C"` in C++ headers). Without a stable ABI it is impossible to know in what register/stack locations to pass arguments and this will lead to endless segmentation faults.

We could potentially support C++ headers which have `extern "C"` on every symbol, but we will never support C++ headers.

dcharkes (approved):

README.md
mannprerak2: Done.

PR #215, opened 2021-05-14
Added support for generating Unions

### LINKED ISSUE #214: Generating Unions
Hi,
I apologise if there is already an issue for this, but either Google isn't being helpful, or I don't understand enough of the terminology to know what I'm looking at.

I'm in the process of trying to write bindings for [SDL](https://github.com/chrisnorman7/dart_sdl), and I'm running into issues.

SDL uses a union [SDL_Event](https://wiki.libsdl.org/SDL_Event), and obviously ffigen isn't generating it.

From my research, it seems as though Union support is coming, but certainly hasn't filtered down to 2.13 as of yet. Is this true? Is there a way I can test the latest ffigen if I switch to Flutter Master?

Thanks in advance.

mannprerak2:
Closes #214 
- Added support for generating unions.
- Added and updated tests.
- Update version, changelog, readme


lib/src/header_parser/sub_parsers/compounddecl_parser.dart
```diff
@@ -0,0 +1,303 @@
+// Copyright (c) 2020, the Dart project authors. Please see the AUTHORS file
```
dcharkes, line 1: 2021 (I guess that might be struct_parser but git not picking up the rename).

lib/src/header_parser/utils.dart
```diff
@@ -326,7 +326,8 @@ class Macro {
 /// Tracks if a binding is 'seen' or not.
 class BindingsIndex {
   // Tracks if bindings are already seen, Map key is USR obtained from libclang.
-  final Map<String, Struc> _structs = {};
+  final Map<String, Compound> _structs = {};
```
dcharkes, line 329: Don't we have subclasses Struc and Union for Compounds?

dcharkes (approved):
Good stuff! 🚚

lib/src/header_parser/sub_parsers/compounddecl_parser.dart
mannprerak2: Yes, it is. Since structs and unions were so similar, it seemed best to not have separate parsers for them.

lib/src/header_parser/utils.dart
mannprerak2: Fixed this, it makes more sense to use struc/union here. I've added a cast operator to the `addStructToSeen` function to fix this.

lib/src/code_generator/compound.dart
```diff
@@ -0,0 +1,182 @@
...
+  /// Marker for if a struct definition is complete.
+  ///
+  /// A function can be safely pass this struct by value if it's complete.
+  bool isInComplete;
+
+  List<Member> members;
+
+  bool get isOpaque => members.isEmpty;
+
+  /// Value for `@Packed(X)` annotation. Can be null(no packing), 1, 2, 4, 8, 16.
```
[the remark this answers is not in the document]
mannprerak2: We probably shouldn't move `pack` to `Struc` class, that will require a few more if/else statements in compound Parser.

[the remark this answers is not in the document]
dcharkes: I have no strong opinion about this, if you think this is clearer, lets go for it!

dcharkes (commented):
CHANGELOG.md?

Version `3.1.0-dev.0`? It's backwards compatible except for the version-bump to Dart 2.14 when we flip it back to stable. And we can release 3.0.x bugfixes if we need to in the mean time.

lib/src/code_generator/compound.dart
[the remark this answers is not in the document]
mannprerak2: Great! Let's keep it intact.

mannprerak2:
Added changelog. 
> It's backwards compatible except for the version-bump to Dart 2.14

Would we need to publish `4.0.0` for dart `2.14` when it lands in stable? In that case we probably should use `4.0.0-dev.0`.

dcharkes:
> Would we need to publish `4.0.0` for dart `2.14` when it lands in stable? In that case we probably should use `4.0.0-dev.0`.

No, it would be fine to publish `3.1.0` for dart `2.14` as long as we satisfy the following constraints:

* We don't change the public Dart API in `lib/*.dart` in a non-breaking way.
* We don't change the command line interface of `bin/*.dart` in a non-breaking way.

Non-breaking with API is defined as: if behavior observably changes. However, with the exception of thrown `Error`s. So you could be throwing an `UnimplementedError` in `3.0.0` and have it implemented in `3.1.0`.

In the same way, we refuse the generate unions in `3.0.0` (analogous to throwing `Error`s in the API), and we implement it in `3.1.0`.

There is some other things we could think about, see https://github.com/dart-lang/ffigen/issues/219.

PR #225, opened 2021-05-27
Add `structs` and `enums`'s configurations in the table

Sunbreak:


mannprerak2 (commented):
I don't really see why this was necessary, but I guess there is no harm in doing this.

Can you move the comments to their own previous line? (Similar to other entries in this cell)
That way the 2nd column gets more space.

mannprerak2 (approved):

Sunbreak:
> I don't really see why this was necessary, but I guess there is no harm adding this.

I was not sure whether `struct`'s `rename` is available because `enum`'s `rename` is not. It took me some time to find the configuration in the FAQ

mannprerak2:
@Sunbreak enum rename is also available.

Sunbreak:
> @Sunbreak enum rename is also available.

I test again and `enum`'s `rename` works. Thanks bro

I'll update the PR with that

dcharkes (approved):

dcharkes:
Thanks @Sunbreak !

Issue #226, opened 2021-05-28
[Feature] Config to expose native-type/dart-type of functions


Sunbreak:
# Background

[My `getNativeSystemInfo`](https://github.com/Sunbreak/dynamic_load.trial/blob/552f0d23c41f97137f5174d5c36024c5cff53471/example/GetNativeSystemInfo.dart#L34-L49) is inspired by [`win32` example](https://github.com/timsneath/win32/blob/main/example/dynamic_load.dart)

```dart
typedef getNativeSystemInfoNative = Void Function(
    Pointer<SYSTEM_INFO> lpSystemInfo);
typedef getNativeSystemInfoDart = void Function(
    Pointer<SYSTEM_INFO> lpSystemInfo);

void getNativeSystemInfo(Pointer<SYSTEM_INFO> lpSystemInfo) {
  final pGetNativeSystemInfo =
      getFuncAddress('kernel32.dll', 'GetNativeSystemInfo');

  final funcGetNativeSystemInfo =
      Pointer<NativeFunction<getNativeSystemInfoNative>>.fromAddress(
              pGetNativeSystemInfo)
          .asFunction<getNativeSystemInfoDart>();

  funcGetNativeSystemInfo(lpSystemInfo);
}
```

And I'd like to call `NtQuerySystemInformation`: https://github.com/Sunbreak/dynamic_load.trial/blob/552f0d23c41f97137f5174d5c36024c5cff53471/example/NtQuerySystemInformation.dart

```dart
    final funcNtQuerySystemInformation =
        Pointer<NativeFunction<c_NtQuerySystemInformation>>.fromAddress(
                pNtQuerySystemInformation)
            .asFunction<dart_NtQuerySystemInformation>();
```

# Problem

`pub run ffigen --config example\ntdll\winternl.yaml` results in private `typedef`s

```dart
typedef _c_NtQuerySystemInformation = ffi.Int32 Function(
  ffi.Int32 SystemInformationClass,
  ffi.Pointer<ffi.Void> SystemInformation,
  ffi.Uint32 SystemInformationLength,
  ffi.Pointer<ffi.Uint32> ReturnLength,
);

typedef _dart_NtQuerySystemInformation = int Function(
  int SystemInformationClass,
  ffi.Pointer<ffi.Void> SystemInformation,
  int SystemInformationLength,
  ffi.Pointer<ffi.Uint32> ReturnLength,
);
```

# Request

Add config to exposure the meta type of functions

```yaml
functions:
[...]

PR #230, opened 2021-06-03
Move from package:pedantic to pacakge:lints

### LINKED ISSUE #229: Move from package:pedantic to package:lints
We've got a new recommended set of lints for Dart packages. https://pub.dev/packages/lint

mannprerak2:
Closes #229 
- Moved from `pacakge:pedantic/analysis_options.yaml` to `package:lints/recommended.yaml`
- Function pointer is now named as `<func>Ptr` (earlier: `<func>_ptr`)
- Name collisions are now resolve by suffixing `<int>` (earlier: `_<int>`)
- Unnamed anonymous structs are now named as `UnnamedStruct<int>` (earlier: `unnamedStruct_<int>`)
- Updated version, changelog


mannprerak2:
@dcharkes, I hope you meant [`package:lints`](https://pub.dev/packages/lints) and not [`package:lint` ](https://pub.dev/packages/lint)?

dcharkes:
> @dcharkes, I hope you meant [`package:lints`](https://pub.dev/packages/lints) and not [`package:lint` ](https://pub.dev/packages/lint)?

Uh, yes! Thanks! 🙈

analysis_options.yaml
```diff
@@ -2,17 +2,19 @@
 # for details. All rights reserved. Use of this source code is governed by a
 # BSD-style license that can be found in the LICENSE file.
 
-include: package:pedantic/analysis_options.yaml
+include: package:lints/recommended.yaml
 
 analyzer:
+  exclude: ['**bindings.dart'] # Exclude all generated bindings.
```
dcharkes, line 8: Can we generate `// ignore_for_file: unused_import, unused_local_variable` in the files instead? So that we know how precisely we are violating the new lints?

We should also explore if some of the lints can be easily satisfied by changing the generator slightly.

mannprerak2: > We should also explore if some of the lints can be easily satisfied by changing the generator slightly.

I am only seeing 2 lint errors -
1. lowerCamelCase => Happens because of how we name the function pointer => using suffix `_ptr`.
2. upperCamelCase => Happens because of how we resolve name conflicts => using suffix `_<int>`.

Should we remove underscore from the suffix, and use => `Ptr` and `<Int>` instead?

dcharkes: Let's do that.

Of course people will still get errors if the C names themselves have underscores (for example `JSON_Version`), but at least they can then make it pass the lints by renaming them if they would want to.

mannprerak2: Done. (The changes have been highlighted in the top comment)

The native type (if exposed using `symbol-address`) is still generated as `Native_<func>` which is not camel case. But I think we can fix this with #226

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 4.0.0-dev.1
+- Use camelCase to name internal variables.
```
dcharkes, line 2: Write the change log from the perspective of our end-users use cases. For example:

- This package now targets `package:lints` for the generated code.
  The generated code uses C symbol names as is.
  Use either `// ignore_for_file: lintRule1, lintRule2` in the preamble, or rename the symbols to make `package:lints` happy.

Which are the most common lint rules that users will have to ignore in the preamble?

dcharkes (approved):
lgtm!

CHANGELOG.md
mannprerak2: Updated changelog.

> Which are the most common lint rules that users will have to ignore in the preamble?

Considering how C libraries are: 
- `camel_case_types, non_constant_identifier_names`. (Most libraries use underscores rather than camel case.)
- `unused_element`. (For C declarations starting with `_`)

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# 4.0.0-dev.1
+- This package now targets package:lints for the generated code. The generated
+code uses C symbol names as is. Use either `// ignore_for_file: lintRule1, lintRule2`
+in the `preamble`, or rename the symbols to make package:lints happy.
+- Function pointers are now named as `_<func>Ptr` instead of `_<func>_ptr`.
```
dcharkes, line 5: This is not visible in the public API right? I think we can omit it.

README.md
```diff
@@ -381,9 +381,7 @@ use-dart-handle: true
 
 ```yaml
 preamble: |
-  /// AUTO GENERATED FILE, DO NOT EDIT.
-  ///
-  /// Generated by `package:ffigen`.
```
dcharkes, line 386: Probably also keep this. It's a good habit for people to do this.

CHANGELOG.md
mannprerak2: Done.

README.md
mannprerak2: Actually, we are writing this for every generated file, and the user cannot remove this.
So using this example for the preamble was _kind of_ misleading.
https://github.com/dart-lang/ffigen/blob/98b2d4ff29a8f054dce52a30b909f423d4cbd567/lib/src/code_generator/writer.dart#L159-L161

dcharkes: Awesome!

PR #254, opened 2021-10-21
Exclude all on empty include list

lukaslihotzki:
When setting `include: []`, I expected nothing will be included. When I saw everything was included anyway, I thought that null and empty list probably are processed the same and the next try `exclude: [".*"]` worked.

Anyway, I think the correct behavior for `include: []` is include nothing. In my use case, I explicitly allowlist every item and `include: []` is more intuitive for the categories without any items.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 4.1.2
+- Added fix for empty include list to exclude all
+
 # 4.1.1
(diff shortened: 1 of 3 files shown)


lukaslihotzki:
@dcharkes I have added the version bump and the changelog.

dcharkes (approved):

PR #261, opened 2021-12-26
Support for ABI-specific integers

mannprerak2:
Closes #7.
This PR lays the groundwork for the generation of ABI-specific integers.

- Added 2 new configs - `library-imports` and `type-map`
- Removed config keys - `size-map` and `typedef-map`.
- `type-map` can target native types, typedefs, structs and unions.
- Update readme, changelog, version


mannprerak2:
@dcharkes This PR currently does not generate or use any ABI-specific integers by default but it's left for the user to do by themselves via the given config. I think it'd better to wait till `package:ffi` has defined these ABI-specific integers before ffigen uses them by default.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# 5.0.0-dev.1
+- Added support for generating ABI Specific integers.
+- Breaking: removed config keys - `size-map` and `typedef-map`.
```
dcharkes, line 3: Should we have a deprecation period, so that people can migrate while still supporting the old keys but give warnings?

dcharkes:
> This PR currently does not generate or use any ABI-specific integers by default but it's left for the user to do by themselves via the given config.

When there's a Dart dev release with it, I can make a dev release for `package:ffi`.

CHANGELOG.md
mannprerak2: I think it's not really necessary in light of the following
 - If the warnings are ignored these keys may override the default ABI specific integers (e.g for size_t) which will not be caught untill runtime on specific platforms.
- The size-map key will probably still need to be removed because of how it sort of clashes with the defaults.
- The lower SDK constraint probably needs to be updated to 2.16 by users.
- package:ffi will also now be mandatory for users as a dependency.
- The users will still be warned about having these old config keys, so it's not like this will go unnoticed.

So i guess a 5.0.0 makes more sense here(which allows us to add breaking changes).

mannprerak2:
> When there's a Dart dev release with it, I can make a dev release for package:ffi.

I'm currently using the 2.16.0-dev.124 release which I guess supports Abi specific integers. So we can possibly hook up package:ffi in this PR if you want.

dcharkes:
I'll make a pre-release once reviewed:

* https://github.com/dart-lang/ffi/pull/128

lib/src/code_generator/imports.dart
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) 2021, the Dart project authors. Please see the AUTHORS file
```
dcharkes, line 1: nit this and other files: 2022 when we're going to squash and merge.

lib/src/code_generator/imports.dart
```diff
@@ -0,0 +1,51 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Represents a library import which will be written as an import in the
```
dcharkes, line 5:
```dart
/// A library import which will be written as an import in the
/// generated file.
```

https://dart.dev/guides/language/effective-dart/documentation#prefer-starting-library-or-type-comments-with-noun-phrases

lib/src/code_generator/imports.dart
```diff
@@ -0,0 +1,51 @@
...
+  @override
+  bool operator ==(other) {
+    return other is LibraryImport && name == other.name;
+  }
+
+  @override
+  int get hashCode => name.hashCode;
+}
+
+/// Represents an imported type.
```
dcharkes, line 25:
```dart
/// An imported type which will be used in the generated code.
```

lib/src/code_generator/imports.dart
```diff
@@ -0,0 +1,51 @@
...
+
+final voidType = ImportedType(ffiImport, 'Void', 'void');
+
+final unsignedCharType = ImportedType(ffiImport, 'Uint8', 'int');
+final charType = ImportedType(ffiImport, 'Int8', 'int');
+final unsignedShortType = ImportedType(ffiImport, 'Uint16', 'int');
+final shortType = ImportedType(ffiImport, 'Int16', 'int');
+final unsignedIntType = ImportedType(ffiImport, 'Uint32', 'int');
+final intType = ImportedType(ffiImport, 'Int32', 'int');
+final unsignedLongType = ImportedType(ffiImport, 'Uint64', 'int');
```
dcharkes, line 45: This variable name is incorrect.

`(unsigned) long` is not (always) `(u)int64_t`. https://dart-review.googlesource.com/c/sdk/+/221501/28/tests/ffi/abi_specific_ints.dart#76

You can play around with https://godbolt.org/z/ee8MTjY8Y if you want to inspect the sizes. You can select the compiler (win/linux x64/x86) in the dropdown and see the size as an int in the generated assembly.

lib/src/header_parser/type_extractor/cxtypekindmap.dart
```diff
@@ -27,6 +28,22 @@ var cxTypeKindToSupportedNativeTypes = <int, SupportedNativeType>{
...
 
+var cxTypeKindToImportedTypes = <String, ImportedType>{
+  'void': voidType,
+  'unsigned char': unsignedCharType,
+  'char': charType,
+  'unsigned short': unsignedShortType,
+  'short': shortType,
+  'unsigned int': unsignedIntType,
+  'int': intType,
+  'unsigned long': unsignedLongType,
```
dcharkes, line 39: long and unsigned long do not have a fixed size type on the various architectures

(unsigned) char, (unsigned) short, (unsigned) int, and (unsigned) long long do. However, for these we should probably also target the typedefs in `package:ffi` for clarity. (The C standard says at least 8 bytes for long long, so we could even have platforms in the future which decide for another size. In that case we can just update `package:ffi` and don't have to update `package:ffigen`.)

lib/src/strings.dart
```diff
@@ -110,6 +107,20 @@ const sizemap_native_mapping = <String, int>{
   Enum: clang.CXTypeKind.CXType_Enum
 };
 
+// Library imports.
+const libraryImports = 'library-imports';
+
+final predefinedLibraryImports = {
+  ffiImport.name: ffiImport,
+  // ffiPkgImport.name: ffiPkgImport // TODO: add this import.
```
dcharkes, line 115: nit: left TODO without issue number

pubspec.yaml
```diff
@@ -3,12 +3,12 @@
 # BSD-style license that can be found in the LICENSE file.
 
 name: ffigen
-version: 4.1.2
+version: 5.0.0-dev.1
```
dcharkes, line 6: nit: I think we can start counting with dev.0

test/example_tests/libclang_example_test.dart
```diff
@@ -43,9 +43,25 @@ ${strings.enums}:
   ${strings.include}:
     - 'CXTypeKind'
     - 'CXGlobalOptFlags'
-${strings.typedefmap}:
-    'size_t': 'IntPtr'
-    'time_t': 'Int64'
+${strings.libraryImports}:
+  custom_import: 'custom_import.dart'
+${strings.typeMap}:
+  'typedef size_t':
```
dcharkes, line 49: why is `typedef ` part of the key?

dcharkes (approved):
Happy new year! 🧨 🎉  

LGTM with comments addressed.

I think we should only map `int8_t` etc to `dart:ffi` types and target `typedef`s or `AbiSpecificInteger`s in `package:ffi` for all other types (including `char`, `int`, etc.)

test/example_tests/libclang_example_test.dart
mannprerak2: I wanted to make `type-map` a generic way to override any import whatsoever. That means we need to be able to target structs, unions, typedefs and finally native types. Since C allows these to be named exactly the same as they are under different namespace, we need to follow suit.

The typemap in key signifies that only typedefs are to be matched. This has been documented in the configurations section of readme.

An alternative way could be to add a nested key for namespace.

lib/src/header_parser/type_extractor/cxtypekindmap.dart
mannprerak2: Those were the defaults before this PR. Will update these once package:ffi has a dev release.

example/libclang-example/custom_import.dart
```diff
@@ -0,0 +1,63 @@
...
+  Abi.macosX64: Uint64(),
+  Abi.windowsArm64: Uint32(),
+  Abi.windowsIA32: Uint32(),
+  Abi.windowsX64: Uint32(),
+})
+class UnsignedLong extends AbiSpecificInteger {
+  const UnsignedLong();
+}
+
+class CXCursorSetImpl extends Opaque {}
```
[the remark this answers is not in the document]
mannprerak2: This is on purpose, since we use type-map to use our custom definition of this struct.

pubspec.yaml
mannprerak2: Ofcourse 🙈 and Happy New year 🎉

.github/workflows/test-package.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        sdk: [2.14.0]
+        sdk: [2.16.0-124.0.dev] # TODO: revert.
```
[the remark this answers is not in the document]
mannprerak2: #262

lib/src/strings.dart
mannprerak2: This TODO will be removed once we get a dev `package:ffi`.

example/libclang-example/generated_bindings.dart
```diff
@@ -9282,7 +9283,7 @@ typedef DartClang_Cursor_getTranslationUnit = CXTranslationUnit Function(
 class CXCursorSetImpl extends ffi.Opaque {}
```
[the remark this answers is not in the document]
mannprerak2: On second thought, I am not sure if we shouldn't generate the custom import.

The user has control over what declaration to include/exclude. Currently, if the user has excluded a declaration, we will not generate it. The custom import will be used instead.

But if we force exclude anything that is inside a custom import. We will never generate it even if the user specifies it.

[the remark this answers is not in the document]
dcharkes: I was thinking that we either:

1. include it from a custom import
2. or generate it

But don't do both and generate an unused declaration.

> Currently, if the user has excluded a declaration, we will not generate it.
In other words: currently, if the user does not exclude something but includes a header file, we generate it.

And currently, we only generate declarations that are (transitively) reachable from other files. We don't generate things that are not used at all.

I can see how these things conflict.

We need to come up with a reasonable way to handle 'shadowing'.

I'd say that shadowing would be: if we have already defined a type in an import, do not regenerate it, unless it is specifically included with an include.

(I just realized that us having imports will also start allowing us to generate bindings that import other generated bindings. 🎉  And it would be also relevant for that. In that case it might even refer to the same sources in C.)

What do you think?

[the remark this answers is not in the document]
mannprerak2: How about adding a custom key for shadow, with default as true. Which can be overriden.
```yaml
type-map:
  'struct MyStruct':
    'lib': 'custom_lib'
    'c-type': 'CustomType'
    'dart-type': 'CustomType'
    'shadow': false
```


Generated bindings using other generated bindings is exactly what I had in mind 🎉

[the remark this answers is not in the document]
dcharkes: > How about adding a custom key for shadow

Hm, if the key is set to true, then the declarations are not generated, but that is not visible in the part where you specify your includes/excludes on those declarations.
And, if the key is set to false, then you might get weird results when the file you're having a custom import in changes and now defines it. You might not recognize that it's shadowed.

I was thinking more of a general shadowing behavior rather than per item:

`type-map-shadowing: prefer-type-map (default) | prefer-local | error`.

With `error`: let the users resolve it with the include/exclude of the main file and the list of imports through the type-map.
With `prefer-type-map`: the type-map always shadows local definitions and we don't generate local definitions
With `prefer-local`: local definitions win out. (Although, I cannot come up with a use case for this.)

Though, this line of thinking does not distinguish between declarations in the `entry-points` and `include-directives`.

I think the default behavior for a conflict between the `type-map` and non-excluded definitions in `entry-points` should be `error`. While the default behavior for a conflict between the `type-map` and `include-directives` should be that the `type-map` wins.

If we have:

* `a.h` defining `struct A`
* `b.h` including `a.h` and using `struct A` and defining a `struct B`.
* `c.h` including `b.h` and defining a function with `struct B`.
* `d.h` also defining `struct A` in the same way as `a.h`.

And then we have two `ffigen` configs:

1. generating for `b.h` with `a.h` in its include directives. (and outputting also a yaml file with a type-map? 😄 )
2. generating for `c.h` with both other files in its include directives and having a type-map for `struct A` and `struct B`.
3. generating for `d.h` with type-map for `struct A`

In (1) we generate `struct A` because it is not defined in `b.h`.
In (2) we don't generate `struct A` and `B` because it's already in the `type-map`.
In (3) we get an error, because we would get conflicting definitions. The user has to not include the local `struct A`.

If we go this route we don't need a config key at all, but some documentation! 😸 

Do you envision use cases in which this would not work?

test/example_tests/libclang_example_test.dart
dcharkes: > An alternative way to implement this could be to add a nested key for namespace.

This feels cleaner to me. It also prevents people from having to repeat the namespace every definition.

example/libclang-example/generated_bindings.dart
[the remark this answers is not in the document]
mannprerak2: This honesty seems too complicated to me. Infact there's no good reason for us to error out and make the user exclude declarations.
The only use case I see for prefer-local would be to sort of compare the generated bindings to see if the struct definitions are same, and then remove them.
Perhaps we can just have a global key to exclude type-map declarations from being generated. A  type-map level key like shadow (from earlier suggestion) would be too verbose.

Also the entry-points are only a starting point for parsing. Whether or not we include a declaration depends on what matched by include-directives.

test/example_tests/libclang_example_test.dart
mannprerak2: So do you mean something like this -

```yaml
type-map:
  typedef:
    'size_t':
      lib: pkg_ffi
      c-type: SIze_t
      dart-type: int
```

OR

```yaml
type-map:
  'size_t:
    lib: pkg_ffi
    c-type: Size_t
    dart-type: int
    namespace: typedef
```

dcharkes: The first one. 👍

example/libclang-example/generated_bindings.dart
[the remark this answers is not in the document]
dcharkes: > Also the entry-points are only a starting point for parsing. Whether or not we include a declaration depends on what matched by include-directives.

Ah, I misremembered that then. Sorry for the confusion.

Let's just implement something for now, so that we can have support for the `type-map`. And then we can always refine that (with a breaking change and major version bump) if we run into use cases which we didn't foresee.

test/example_tests/libclang_example_test.dart
mannprerak2: Done. Also updated readme.

lib/src/header_parser/type_extractor/cxtypekindmap.dart
mannprerak2: Updated

lib/src/code_generator/imports.dart
mannprerak2: Updated to use types from package:ffi.

mannprerak2:
@dcharkes I've updated the code to use `package:ffi` everywhere, by using a git dependency. But this seems to have broken the tests on CI. Would update the PR to point to the dev version once it's released.

README.md
```diff
@@ -431,43 +433,54 @@ preamble: |
...
-  long: 8
-  unsigned long: 8
-  long long: 8
-  unsigned long long: 8
-  enum: 4
+type-map:
+  'native-types': # Targets native types.
+    'char':
+      'lib': 'ffi' # predefined import.
+      'c-type': 'Int8'
```
dcharkes, line 461: 'c-type': 'Char' ?

README.md
```diff
@@ -431,43 +433,54 @@ preamble: |
...
+      # For native-types dart-type can be be int, double or float
+      # but same otherwise.
+      'dart-type': 'int'
+    'int':
+      'lib': 'custom_lib'
+      'c-type': 'AbiSpecificInt'
```
dcharkes, line 467: Maybe also CustomType[0-9]*, to clarify that this one is not Special compared to the others.

example/c_json/pubspec.yaml
```diff
@@ -26,8 +29,12 @@ ffigen:
     include-directives:
       - '**cJSON.h'
   comments: false
-  typedef-map:
-    'size_t': 'IntPtr'
+  type-map:
+    typedefs:
+      'size_t':
+        lib: 'ffi'
+        c-type: 'IntPtr'
```
dcharkes, line 36: c-type: 'Size'

example/libclang-example/pubspec.yaml
```diff
@@ -51,25 +56,30 @@ ffigen:
...
 
   # These can be used to map any typedef to a specific native type.
-  typedef-map:
-    'size_t': 'IntPtr'
-    'time_t': 'Int64'
+  type-map:
+    typedefs:
+      'size_t':
+        lib: 'custom_import'
+        c-type: 'UintPtr'
```
dcharkes, line 67: ctype: 'Size'

lib/src/code_generator/imports.dart
```diff
@@ -0,0 +1,54 @@
...
+
+  ImportedType(this.libraryImport, this.cType, this.dartType);
+}
+
+final ffiImport = LibraryImport('ffi', 'dart:ffi');
+final ffiPkgImport = LibraryImport('pkg_ffi', 'package:ffi/ffi.dart');
+
+final voidType = ImportedType(ffiImport, 'Void', 'void');
+
+final unsignedCharType = ImportedType(ffiPkgImport, 'UnsignedChar', 'int');
```
dcharkes, line 36: Turns out that `char`, `signed char` and `unsigned char` are different. (I updated the package:ffi PR.)

lib/src/header_parser/clang_bindings/clang_bindings.dart
```diff
@@ -9,6 +9,7 @@
 //
 // Generated by `package:ffigen`.
 import 'dart:ffi' as ffi;
+import 'package:ffi/ffi.dart' as pkg_ffi;
```
dcharkes, line 12: Woohoo our self-hosted bindings now run on all target platforms 🏖️

lib/src/header_parser/type_extractor/cxtypekindmap.dart
```diff
@@ -27,6 +28,22 @@ var cxTypeKindToSupportedNativeTypes = <int, SupportedNativeType>{
   clang.CXTypeKind.CXType_Char_U: SupportedNativeType.Uint8,
 };
 
+var cxTypeKindToImportedTypes = <String, ImportedType>{
+  'void': voidType,
+  'unsigned char': unsignedCharType,
+  'char': charType,
```
dcharkes, line 34: += 'signed char' (I updated the package:ffi PR)

test/large_integration_tests/large_test.dart
```diff
@@ -37,9 +37,16 @@ ${strings.headers}:
...
     - '**FataErrorHandler.h'
     - '**Index.h'
-${strings.typedefmap}:
-  'size_t': 'IntPtr'
-  'time_t': 'Int64'
+${strings.typeMap}:
+  ${strings.typeMapTypedefs}:
+    'size_t':
+      lib: 'ffi'
+      c-type: 'IntPtr'
```
dcharkes, line 44: c-type: 'Size'

test/large_integration_tests/large_test.dart
```diff
@@ -64,8 +71,12 @@ ${strings.headers}:
     - third_party/cjson_library/cJSON.h
   ${strings.includeDirectives}:
     - '**cJSON.h'
-${strings.typedefmap}:
-  'size_t': 'IntPtr'
+${strings.typeMap}:
+  ${strings.typeMapTypedefs}:
+    'size_t':
+      lib: 'ffi'
+      c-type: 'IntPtr'
```
dcharkes, line 78: c-type: 'Size'

dcharkes (approved):
lgtm!

I'll try to get the `package:ffi` PR landed soon as well :)

FYI: https://dart-review.googlesource.com/c/sdk/+/228541 We're adding the C types to `dart:ffi` in Dart 2.17 (too late for 2.16) so that we can use them in `@FfiNative` annotations  (which are used in places where we cannot import `package:ffi`). I'll update `package:ffi` to just export the `dart:ffi` definitions when Dart 2.17 comes around.

README.md
dcharkes: I just meant naming it CustomType4 to be consistent with the other custom types in the readme.

dcharkes:
https://pub.dev/packages/ffi/versions/1.2.0-dev.0/changelog is here!

(Note I dropped ssize_t and off_t because they're only part of the posix extension not of C themselves. If we turn out to need these, we can make an extra PR to package:ffi to include some posix types.)

lib/src/header_parser/type_extractor/cxtypekindmap.dart
mannprerak2: Done

dcharkes:
Thanks @mannprerak2! 🥇 

https://pub.dev/packages/ffigen/versions/5.0.0-dev.0

PR #278, opened 2022-02-24
Add default location for LLVM on M1 Mac

KyleFin:
I've verified on my M1 Macbook Pro that `brew install llvm` installs llvm in /opt/homebrew/opt/llvm/lib/ (as the [Homebrew documentation](https://docs.brew.sh/Installation) explains:  `/usr/local for macOS Intel, /opt/homebrew for Apple Silicon`)

Currently this path can be provided with symlinks or in ffigen's config with llvm-path, but it would be great to check the default brew install location by default.

### DIFF
--- a/lib/src/strings.dart
+++ b/lib/src/strings.dart
@@ -186,6 +186,7 @@ const windowsDylibLocations = [
 ];
 const macOsDylibLocations = [
   '/usr/local/opt/llvm/lib/',
+  '/opt/homebrew/opt/llvm/lib/',
 ];
 
 // Writen doubles.
--- a/lib/src/strings.dart
+++ b/lib/src/strings.dart
@@ -174,20 +174,20 @@ const libclang_dylib_macos = 'libclang.dylib';
 const libclang_dylib_windows = 'libclang.dll';
 
 // Dynamic library default locations.
-const linuxDylibLocations = [
+const linuxDylibLocations = {
   '/usr/lib/llvm-9/lib/',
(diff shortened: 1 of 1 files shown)


dcharkes (approved):
Thanks @KyleFin!

Please also sign the CLA.

KyleFin:
Done. I think it's ready to merge. Thanks @dcharkes!

lib/src/strings.dart
```diff
@@ -186,6 +186,7 @@ const windowsDylibLocations = [
 ];
 const macOsDylibLocations = [
```
kevmoo, line 187: nit - this could be a `Set` 😄

KyleFin: Done. Looks safe to me for windows and linux too:
https://github.com/dart-lang/ffigen/search?q=macOsDylibLocations
https://github.com/dart-lang/ffigen/search?q=windowsDylibLocations
https://github.com/dart-lang/ffigen/search?q=linuxDylibLocations

dcharkes: A [LinkedHashSet](https://api.dart.dev/stable/2.13.4/dart-collection/LinkedHashSet-class.html), which a const set is. Because we would not want to make the iteration order undefined and make llvm resolution non-deterministic.

mannprerak2: Does it really make a difference though? Since we only need to iterate, an array should be the same as a LinkedHashSet in that regard.

PR #281, opened 2022-03-02
First bits of Objective C support

liamappelbe:
First bits of objective C support

- Add a language flag to the config
- Enable some objective C flags in the clang invocation, when language is set to objective C
- Support some basic objective C types that don't require any special parsing: `BOOL`, `id`, `SEL`, `NSObject*`, `Class*`, and block pointers

#279


test/large_integration_tests/_expected_libclang_bindings.dart
```diff
@@ -7462,6 +7524,9 @@ typedef CXCursorVisitor = ffi.Pointer<
 /// various callbacks and visitors.
 typedef CXClientData = ffi.Pointer<ffi.Void>;
 
+/// Visitor invoked for each cursor found by a traversal.
+typedef CXCursorVisitorBlock = ffi.Pointer<ffi.Opaque>;
```
liamappelbe, line 7528: Interesting. Looks like libclang will parse this and return a block pointer, even when objective c mode is not turned on.

https://intel.github.io/llvm-docs/clang_doxygen/Index_8h_source.html

liamappelbe:
@dcharkes what's up with the analysis errors? What should I be passing to the native function, instead of Pointer<Opaque>?

dcharkes:
> @dcharkes what's up with the analysis errors? What should I be passing to the native function, instead of Pointer?

Does this also happen with `Pointer<Foo>` and `Foo extends Opaque` instead of `Pointer<Opaque>`. I don't see any other reason. I can spin it up locally if that's not the culprit.

liamappelbe:
> > @dcharkes what's up with the analysis errors? What should I be passing to the native function, instead of Pointer?
> 
> Does this also happen with `Pointer<Foo>` and `Foo extends Opaque` instead of `Pointer<Opaque>`. I don't see any other reason. I can spin it up locally if that's not the culprit.

Yep, that fixed it.

test/large_integration_tests/_expected_libclang_bindings.dart
liamappelbe: Seems like it only does this on mac. I'll make the objective C stuff in extractor.dart conditional on the language flag.

lib/src/strings.dart
```diff
@@ -29,6 +29,19 @@ String get dynamicLibParentName => Platform.isWindows ? 'bin' : 'lib';
...
 const output = 'output';
 
+const language = 'language';
+
+// String mappings for the Language enum.
+const langC = 'c';
+const langObjC = 'objc';
+
+// Clang command line args for Objective C.
+const clangLangObjC = ['-x', 'objective-c'];
```
mannprerak2, line 39: Should we also pass a language flag when lang is `C`? Not sure if that would make any difference though.

test/header_parser_tests/objc_basic_types.h
```diff
@@ -0,0 +1,8 @@
+struct Foo {
+  BOOL someBool;
+  id anId;
+  SEL selector;
+  NSObject* object;
+  Class* clazz;
```
mannprerak2, line 6: We should add some larger integration tests and some obj c examples later.

mannprerak2 (approved):
LGTM 👍

test/header_parser_tests/expected_bindings/_expected_objc_basic_types_bindings.dart
```diff
@@ -0,0 +1,210 @@
...
+  NativeLibrary(ffi.DynamicLibrary dynamicLibrary)
+      : _lookup = dynamicLibrary.lookup;
+
+  /// The symbols are looked up with [lookup].
+  NativeLibrary.fromLookup(
+      ffi.Pointer<T> Function<T extends ffi.NativeType>(String symbolName)
+          lookup)
+      : _lookup = lookup;
+
+  int __darwin_check_fd_set_overflow(
```
dcharkes, line 23: Where do these functions come from? I see only the struct in the in the header file.

liamappelbe: parser.dart:70. I have to auto-include some system headers to get it to parse. Some of the internal built in types are defined in there.

mannprerak2: I guess we should be able to exclude these declarations by matching on its source file, see this -
https://github.com/dart-lang/ffigen/blob/dd3e08945a5be158981c3e1e852ef4b6e56c5547/lib/src/header_parser/includer.dart#L72

lib/src/strings.dart
liamappelbe: It defaults to C, so this isn't needed.

test/header_parser_tests/expected_bindings/_expected_objc_basic_types_bindings.dart
mannprerak2: I think Ideally we should, but I guess we cannot if these might be needed by users.

dcharkes: I think we should not generate them until we have a use case in a follow up CL. Otherwise it is not clear what is the mapping between the header file + config yaml file to the output.

dcharkes (approved):
lgtm with comments addressed

test/header_parser_tests/expected_bindings/_expected_objc_basic_types_bindings.dart
liamappelbe: Done

PR #283, opened 2022-03-14
BindingsIndex refactor

### LINKED ISSUE #282: Refactor BindingsIndex
It's currently pretty difficult to trace how `CXType`/`CXCursor` is converted to a `Type`. Some types are cached in `BindingsIndex` and some aren't, and the caching logic is duplicated in a few places. The logic is also split across a few files. This is making it difficult to add Objective C support. It would be cleaner to have a single entrypoint to the conversion, with a single cache.

The design I have in mind is:
- `getCodeGenType` is still the only entrypoint, and takes a `CXType` (we might be able to remove the `pointerReference` param though).
- The first thing this function does is lookup the `CXType` in the type cache (`Map<String, Type>`).
- If not found, the `Type` is created based on the kind and inserted into the cache, but not filled in yet (due to possible self-reference cycles)
- Then the type is filled in using the parsers. This may recursively call `getCodeGenType`, but cycles will be cache hits, so this is ok.

For the cache key, we can use the USR, but this is only present for types that have a declaration. Types that don't have a USR are basic types like ints, or derived types like pointers, so we can just skip the cache and use the existing logic.

liamappelbe:
Unify and simplify how Types are cached and cycles are broken in the type graph.

- All types that need to be cached (those that have a declaration cursor) are cached in `BindingsIndex._declaredTypes`
- All calls to convert a `CXType` to an ffigen `Type` now go through `getCodeGenType`
- All calls to the related sub-parsers (`parseCompoundDeclaration`, `parseEnumDeclaration`, `parseTypedefDeclaration`) also go through `getCodeGenType`
- Those sub-parsers (mostly) don't know about `BindingsIndex`, and don't know about caching or cycle breaking. That's all handled in one place in `getCodeGenType`
- Cycles are broken by dividing type conversion into 2 stages (the only conversion this matters for is compound types):
    - First the `Type` is created, with as many fields filled in as possible, without a recursive call to `getCodeGenType` (everything except the list of members)
    - Next, `getCodeGenType` inserts the type into the cache
    - Finally, the members are filled in. Recursive calls to `getCodeGenType` are allowed here, as they will hit the cache, breaking any cycles.

Fixes #282

### DIFF
--- a/example/libclang-example/generated_bindings.dart
+++ b/example/libclang-example/generated_bindings.dart
@@ -7966,6 +7966,26 @@ typedef NativeClang_getDiagnosticInSet = CXDiagnostic Function(
     CXDiagnosticSet Diags, pkg_ffi.UnsignedInt Index);
 typedef DartClang_getDiagnosticInSet = CXDiagnostic Function(
     CXDiagnosticSet Diags, int Index);
+
+/// Describes the kind of error that occurred (if any) in a call to
+/// \c clang_loadDiagnostics.
+abstract class CXLoadDiag_Error {
+  /// Indicates that no error occurred.
+  static const int CXLoadDiag_None = 0;
+
+  /// Indicates that an unknown error occurred while attempting to
+  /// deserialize diagnostics.
+  static const int CXLoadDiag_Unknown = 1;
+
+  /// Indicates that the file containing the serialized diagnostics
+  /// could not be opened.
+  static const int CXLoadDiag_CannotLoad = 2;
+
+  /// Indicates that the serialized diagnostics file is invalid or
+  /// corrupt.
+  static const int CXLoadDiag_InvalidFile = 3;
+}
+
 typedef NativeClang_loadDiagnostics = CXDiagnosticSet Function(
     ffi.Pointer<pkg_ffi.Char> file,
     ffi.Pointer<ffi.Int32> error,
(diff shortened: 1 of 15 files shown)


liamappelbe:
There's one test still failing, and I need to clean up some debugging code, but I think this is ready for you guys to start looking at.

I think the failing test is because the new system includes some stuff that used to be excluded. I'll dig into this a bit more today.

example/libclang-example/generated_bindings.dart
```diff
@@ -9296,8 +9296,6 @@ typedef NativeClang_Cursor_getTranslationUnit = CXTranslationUnit Function(
 typedef DartClang_Cursor_getTranslationUnit = CXTranslationUnit Function(
     CXCursor arg0);
 
-class CXCursorSetImpl extends ffi.Opaque {}
```
liamappelbe, line 9299: This was one of the reasons libclang_example_test.dart was failing, but it looks like a bug to me. `CXCursorSetImpl` is not used anywhere in this file other than `CXCursorSet` below, which has been explicitly overwritten to use `custom_import.CXCursorSetImpl`. So I think there must have been a bug where this binding was included anyway, and I inadvertently fixed it.

lib/src/header_parser/utils.dart
```diff
@@ -89,14 +89,14 @@ extension CXCursorExt on clang_types.CXCursor {
     return s;
   }
 
-  /// Dispose type using [type.dispose].
```
liamappelbe, line 92: I think this is a mistake in the documentation. Doesn't look like it's necessary (or possible) to dispose a `CXType`

liamappelbe:
Fixed all the tests and cleaned up. Ready for full review.

mannprerak2 (approved):
LGTM 🎉

example/libclang-example/generated_bindings.dart
mannprerak2: That's strange. The struct CXCursorSetImpl is included according to the config. So it should've been generated regardless of it's usage.

Is it possible it got excluded because it is also in the type-map config?

mannprerak2: The current behaviour in type-map is to use the custom-import if it matches, but the actual definitely is still generated if the user includes it in the config.

This might be happening because all Declaration type generations now go through getCodegenType.

@dcharkes if this behaviour is fine, we can perhaps let this be. The catch is that one would need to remove the type-map to generate the original definition.

dcharkes: If we have it in the type map, I cannot come up with a reason for people to want to have the generated one as well.l

If there's a use who does, we can always make a patch again.

dcharkes (approved):

PR #290, opened 2022-03-30
Make type polymorphic

### LINKED ISSUE #284: Clean up: make Type polymorphic
It looks like `Type` is a thin wrapper around a lot of different kinds of type, each with a different implementation class. The `BroadType` says which kind of type it is, and each one has a nullable field containing the implementation. So we end up with a lot of switches and ifs on the `BroadType`, and then a lot of null check `!` operators to access the implementation. For example, if `Type.broadType` is `BroadType.Enum`, it means `Type.enumClass` should be non-null.

A simpler approach might be for `Type` to be an interface, and each of the different kinds of type would just implement this interface. Then the `BroadType` checks would change to `type is EnumClass`, and the switches would just be polymorphic method calls. There is already a base class for all the implementations like `EnumClass`, called `Binding`, so maybe `Type` could be a subclass of `NoLookupBinding`?

liamappelbe:
Switch `Type` from a thin wrapper around a bunch of classes, to being the base class for those classes.

Remove `BroadType`
 - All `BroadType` checks are now `is` checks
 - All `switch` statements using `BroadType` are now polymorphic method calls

Also
 - Fixed a few typos like `Struc` and `isInComplete`
 - Added `BindingType` which is a work around for the fact that Dart doesn't have multiple inheritance. Think of `BindingType` as extending both `NoLookUpBinding` and `Type`.
 - There are a lot of cases where stuff like `Type.importedType(intType)` can just become `intType` because `ImportedType` extends `Type`

Fixes #284


dcharkes:
Drive by comment: nice! That makes the code nicer.

lib/src/header_parser/type_extractor/extractor.dart
```diff
@@ -100,17 +99,17 @@ Type getCodeGenType(
...
         .CXType_ConstantArray: // Primarily used for constant array in struct members.
-      return Type.constantArray(
+      return ConstantArray(
         clang.clang_getNumElements(cxtype),
         clang.clang_getArrayElementType(cxtype).toCodeGenType(),
       );
     case clang_types.CXTypeKind
         .CXType_IncompleteArray: // Primarily used for incomplete array in function parameters.
-      return Type.incompleteArray(
+      return IncompleteArray(
```
mannprerak2, line 108: I think we could have some static functions on Type class to create objects of its subtypes. So instead of doing PointerType() we could still do Type.pointer(). This probably would have reduced the diff a lot.

mannprerak2: Not suggesting a change, but just an observation.

liamappelbe: Yeah, it would have reduced the diff. But I'm not sure what the point of those functions would be in the new system. It seemed simpler to me to just remove them.

mannprerak2 (approved):

dcharkes (approved):
lgtm

PR #292, opened 2022-04-06
Add a warning message to objective C support

liamappelbe:


lib/src/config_provider/spec_utils.dart
```diff
@@ -464,7 +464,12 @@ Language languageExtractor(dynamic value) {
 
 bool languageValidator(List<String> name, dynamic value) {
   if (value is String) {
-    if (value == strings.langC || value == strings.langObjC) {
+    if (value == strings.langC) {
+      return true;
+    }
+    if (value == strings.langObjC) {
+      _logger.warning('Objective C support is EXPERIMENTAL. The API may change '
```
dcharkes, line 471: I'd prefer using `severe` here. That way it will always show up.

dcharkes (approved):
lgtm with comments.

https://github.com/dart-lang/ffigen#configurations should have documentation for the new `language` key, and should document it as EXPERIMENTAL as well.

PR #293, opened 2022-04-07
Add an integration test for objective C

liamappelbe:
Added an integration test for objective C, based off native_test. It's bare bones at the moment, just creating an instance and poking at it. I'll add more elements to the test as I add more features.

Added a test/setup.dart script that wraps the existing build_test_dylib.dart script and the new objective c test's setup script. We can add additional scripts here as we add more languages/integration tests. The wrapper script also takes care of setting up the working directory, so users don't have to cd anywhere.

Also fixed a couple of minor issues I noticed:
 - Analyzer says setters shouldn't bother stating that the return type is `void`, so I removed that from the code gen.
 - I was getting a lot of spurious warnings that the private functions we use for ObjC code gen (eg _objc_msgSend) start with _ so are private. Added an `isInternal` flag to `Binding` so we can hide that warning for internal things.


test/native_objc_test/native_objc_test.dart
```diff
@@ -0,0 +1,62 @@
...
+        rethrow;
+      }
+    });
+
+    test('Interface basics, with Foo', () {
+      final foo1 = Foo.makeFoo(lib, 3.14159);
+      final foo2 = Foo.makeFoo(lib, 2.71828);
+      expect(foo1.intVal, 3);
+      expect(foo2.intVal, 2);
+      expect(foo1.multiply(0, foo2), 8);
```
dcharkes, line 55: cool!

test/setup.dart
```diff
@@ -0,0 +1,36 @@
...
+    throw ProcessException(Platform.executable, args, '$script failed', result);
+  }
+}
+
+Future<void> main() async {
+  await _run('native_test', 'build_test_dylib.dart');
+  if (Platform.isMacOS) {
+    await _run('native_objc_test', 'setup.dart');
+  }
+  print('\nSuccess :)\n');
```
dcharkes, line 35: Please report the generated file so that it's consistent with the other build:

from the log:

```
Run dart test/setup.dart

Running build_test_dylib.dart in /Users/runner/work/ffigen/ffigen/test/native_test
Building Dynamic Library for Native Tests... 

Generated file: native_test.dylib

Running setup.dart in /Users/runner/work/ffigen/ffigen/test/native_objc_test
Building Dynamic Library for Objective C Native Tests...

Success :)
```

dcharkes (approved):
lgtm

PR #308, opened 2022-04-08
Fix handling of void methods

### LINKED ISSUE #301: ObjC void methods
In the AST, methods that return void look malformed, and are being ignored by the parser. Need to handle them correctly.

liamappelbe:
The AST for void methods just doesn't have a return type, so the quick fix is to just default to void when the return type is unset. But this creates problems for property getters and setters.

The AST contains an entry for the property, and also entries for the getter and setter. The getter and setter entries don't have a return type (which makes sense for the setter, but is incorrect for the getter), so I was deduping these by just ignoring all functions that don't have a return type. With the above fix, the parsed methods now have 2 entries for each getter and setter, one that's correctly marked as a property and has the correct return type, and one that's marked as an ordinary method and has the wrong return type. So I added a filter that ignores the malformed duplicate.

Fixes #301


dcharkes (approved):
lgtm

PR #310, opened 2022-04-08
NSString <-> String converters

### LINKED ISSUE #295: User friendly bindings for NSString
Come up with a user friendly API that looks similar to [the official one](https://developer.apple.com/documentation/foundation/nsstring?language=objc) and handles converting to and from Dart strings

liamappelbe:
Add a factory constructor for `NSString`, and a `toString()` method, for converting to and from a Dart `String`.

Fixes #295

### DIFF
--- a/lib/src/code_generator/objc_interface.dart
+++ b/lib/src/code_generator/objc_interface.dart
@@ -106,6 +106,22 @@ class _ObjCBuiltInFunctions {
       func.addDependencies(dependencies);
     }
   }
+
+  void generateNSStringUtils(Writer w, StringBuffer s) {
+    // Generate a constructor that wraps stringWithCString.
+    s.write('  factory NSString(${w.className} _lib, String str) {\n');
+    s.write('    final cstr = str.toNativeUtf8();\n');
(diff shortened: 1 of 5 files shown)


lib/src/code_generator/objc_interface.dart
```diff
@@ -106,6 +106,22 @@ class _ObjCBuiltInFunctions {
       func.addDependencies(dependencies);
     }
   }
+
+  void generateNSStringUtils(Writer w, StringBuffer s) {
+    // Generate a constructor that wraps stringWithCString.
+    s.write('  factory NSString(${w.className} _lib, String str) {\n');
```
dcharkes, line 112: Why not add an extension on `String` with a method `toNSString()` instead?

dcharkes (approved):
lgtm

lib/src/code_generator/objc_interface.dart
liamappelbe: Done. I think it makes sense to keep the constructor though, because this the main way of constructing an NSString, so it should have a constructor.

Issue #280, opened 2022-03-01
How will YAML change as we add more languages?


liamappelbe:
We need to figure out how to adapt the YAML config as we add more languages (eg [ObjC](https://github.com/dart-lang/ffigen/issues/279), Java, Swift, Kotlin).

For the input source files, for Java/Kotlin/Swift we can just look at the file extension, but ObjC also uses .h files. The easiest solution would just be to add an `objc-headers` key, at the same level as `headers`, and I think that would also make sense for the other languages.

The main issue I see is that most of the other top level config keys are either not applicable to some languages (eg `structs`/`typedefs`/`unions`/`macros`), or we'd want them separated by language (eg `functions` might have different naming conventions in different languages, so the regexes would look different).

Should we just add a bunch more top level keys, but with language specific prefixes? Eg `objc-functions`, `java-classes`? Or maybe we should totally restructure the format so that the top level keys so that the language is at the top level:

```
ffigen:
  output:
  c:
    headers:
    functions:
    ...
  objc:
    headers:
    classes:
    ...
```

mannprerak2:
How about adding a top level `lang` key to the config.
```yaml
#  pubspec.yaml
ffigen:
  lang: 'c'
  headers: ...
  functions: ...
```

This still doesn't provide a way to specify multiple configurations in a single file, but that seems rare since bindings would usually be abstracted away as packages. And one can always specify multiple config files in that case. (Each for a different language/binding)

liamappelbe:
@mannprerak2 @dcharkes After Prerak's comment on #313, I think that methods need a different approach. Similar to how member-rename allows renaming a member of a specific struct, I think users are going to want to be able to include/exclude methods of specific interfaces. So having a top level objcMethods config isn't a good idea, as it would apply to all methods in all interfaces. I'm thinking of doing something like this instead:

```yaml
objcInterfaces:
  member-rename:
    'NSString':
      ...
  member-include:
    'NSString':
      - '.*'
  member-exclude:
    'NSString':
      - 'length'
```

So member-rename applies to interface methods the same way it applies to struct members. And member-include/exclude have the same sort of string keys as member-rename, but full include/exclude lists as values.

I might try to land #313 with just the objcInterfaces config key for now, and just remove the objcMethods config, until we figure out the details of how to hand methods.

dcharkes:
That does make a lot of sense indeed. Methods are nested in interfaces/classes in the same way fields are nested in structs.

[...]

PR #313, opened 2022-04-11
Configurable filtering for ObjC interfaces

liamappelbe:
Add configurable filtering for objective C interfaces, reusing the existing `Declaration` infrastructure.

#279 #280


test/header_parser_tests/objc_interface_test.dart
```diff
@@ -29,6 +29,14 @@ ${strings.language}: '${strings.langObjC}'
 ${strings.headers}:
   ${strings.entryPoints}:
     - 'test/header_parser_tests/objc_interface.h'
+${strings.objcInterfaces}:
+  ${strings.include}:
+    - 'Foo'
```
mannprerak2, line 34: Maybe we can add rename config in the tests as well.

lib/src/header_parser/sub_parsers/objcinterfacedecl_parser.dart
```diff
@@ -106,35 +113,46 @@ void _parseProperty(clang_types.CXCursor cursor) {
...
-  );
-  setter.returnType = NativeType(SupportedNativeType.Void);
-  setter.params.add(ObjCMethodParam(fieldType, 'value'));
-  itf.addMethod(setter);
+  final getterName =
+      clang.clang_Cursor_getObjCPropertyGetterName(cursor).toStringAndDispose();
+  if (config.objcMethods.shouldInclude(getterName)) {
+    final getter = ObjCMethod(
+      originalName: getterName,
+      name: config.objcMethods.renameUsingConfig(getterName),
```
mannprerak2, line 121: The declaration config also provides `renameMemberUsingConfig`. Can that be of any use?

In C it is used to rename the members of struct/unions

test/native_objc_test/config.yaml
```diff
@@ -10,6 +10,10 @@ name: NativeObjCLibrary
 description: 'Native Objective C test'
 language: objc
 output: 'test/native_objc_test/native_objc_test_bindings.dart'
+objcInterfaces:
```
dcharkes, line 13: No change to the generated file?

Maybe add an extra one.

test/native_objc_test/config.yaml
```diff
@@ -10,6 +10,10 @@ name: NativeObjCLibrary
...
+  include:
```
dcharkes, line 14: test rename as well?

lib/src/header_parser/sub_parsers/objcinterfacedecl_parser.dart
liamappelbe: Not sure. I separated objcInterfaces and objcMethods because I needed to be able to filter the methods, and there's no way of filtering members. Maybe it would be cleaner to add that though. Then there could just be one objcInterfaces config that we use to filter/rename both the interfaces and their methods

mannprerak2: Yeah, and it would probably be more consistent with structs/unions/enums, wdyt @dcharkes ?

README.md
```diff
@@ -184,12 +184,15 @@ compiler-opts-automatic:
...
+    <td>
+      functions<br><br>structs<br><br>unions<br><br>enums<br><br>
+      unnamed-enums<br><br>macros<br><br>globals<br><br>objcInterfaces
+    </td>
     <td>Filters for declarations.<br><b>Default: all are included.</b><br><br>
     Options -<br>
     - Include/Exclude declarations.<br>
     - Rename declarations.<br>
-    - Rename enum and struct members.<br>
+    - Rename enum and struct members, and ObjC interface methods.<br>
```
dcharkes, line 195: This is no longer part of this PR correct?

dcharkes (commented):
@liamappelbe did you want to land this PR with only `objc-interfaces` include/exclude? (Based on the API discussion in https://github.com/dart-lang/ffigen/issues/280#issuecomment-1097308166)

```
objc-interfaces:
  include:
    - 'Foo'
    - 'StringUtil'
```

liamappelbe:
Yeah, I think it makes sense to land this PR and then do the method filtering as a follow up.

dcharkes (approved):
lgtm to land filtering without renaming.

PR #314, opened 2022-04-12
Add support for readonly and class properties

brianquinlan:
Fixes https://github.com/dart-lang/native/issues/296

### DIFF
--- a/lib/src/code_generator/objc_interface.dart
+++ b/lib/src/code_generator/objc_interface.dart
@@ -225,28 +225,25 @@ class ObjCInterface extends BindingType {
       s.write('  ');
       if (isStatic) {
         s.write('static ');
+        s.write(_getConvertedType(returnType, w, name));
 
         switch (m.kind) {
           case ObjCMethodKind.method:
             // static returnType methodName(NativeLibrary _lib, ...)
-            s.write(_getConvertedType(returnType, w, name));
             s.write(' $methodName');
-            s.write(paramsToString(m.params, isStatic: true));
             break;
           case ObjCMethodKind.propertyGetter:
             // static returnType getMethodName(NativeLibrary _lib)
-            s.write(_getConvertedType(returnType, w, name));
             s.write(' get');
             s.write(methodName[0].toUpperCase() + methodName.substring(1));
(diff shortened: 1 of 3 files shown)


lib/src/code_generator/objc_interface.dart
```diff
@@ -209,35 +221,56 @@ class ObjCInterface extends BindingType {
...
+          case ObjCMethodKind.propertyGetter:
+            // static returnType getMethodName(NativeLibrary _lib)
+            s.write(_getConvertedType(returnType, w, name));
+            s.write(' get');
+            s.write(methodName[0].toUpperCase() + methodName.substring(1));
+            s.write(paramsToString([], isStatic: true));
+            break;
+          case ObjCMethodKind.propertySetter:
+            // static void setMethodName(NativeLibrary _lib, ...)
+            s.write(' void set');
```
liamappelbe, line 245: Nit: This setter should have a `returnType` of void, so you can change this to be consistent with this other cases, and that means you can pull the `s.write(_getConvertedType(returnType, w, name));` out the front where the static keyword is.

Similarly, the `paramsToString` call can be the same for all 3 cases, so can be moved below the switch.

lib/src/code_generator/objc_interface.dart
```diff
@@ -247,8 +280,12 @@ class ObjCInterface extends BindingType {
       }
       s.write('    $selName ??= '
           '${_builtInFunctions.registerName}(_lib, "${m.originalName}");\n');
-      final convertReturn = _needsConverting(returnType);
-      s.write('    ${convertReturn ? 'final _ret = ' : 'return '}');
+      final convertReturn = m.kind != ObjCMethodKind.propertySetter &&
```
liamappelbe, line 283: Maybe this check should be whether the return type is void, rather than checking if its a setter?

liamappelbe (approved):
LGTM, with a couple of nits

brianquinlan:
@liamappelbe I don't have permission to merge, could you press the button for me?

dcharkes:
@brianquinlan I've given you merge permissions.

PR #326, opened 2022-04-15
Improve Type deduping in objc_msgSend

### LINKED ISSUE #303: Cleanup: Improve objc_sendMsg deduping
Currently we generate a lot of objc_sendMsg functions that have the same signature, because the deduping isn't very accurate. This leads to a lot of unnecessary boilerplate in the generated code.

liamappelbe:
This is also important for Blocks, which will be deduped using the same infrastructure. Reduces the amount of boilerplate generated by about 20%.

Fixes #303


lib/src/code_generator/type.dart
```diff
@@ -47,6 +47,12 @@ abstract class Type {
   /// only. This string should not be printed as generated code.
   @override
   String toString();
+
+  /// Cache key used in various places to dedupe Types. By default this is just
+  /// the hash of the Type, but in many cases this does not dedupe sufficiently.
+  /// So Types that may be duplicated should override this to return a more
+  /// specific key. Types that are already deduped don't need to override this.
+  String cacheKey() => hashCode.toRadixString(36);
```
dcharkes, line 55: Maybe document why `toString` as cacheKey doesn't work.

(I presume we could have name collisions?)

dcharkes (approved):

lib/src/code_generator/type.dart
liamappelbe: Yeah. toString was just designed for debugging, so I haven't thought carefully about making it unambiguous.

brianquinlan (commented):
Is there no end-to-end test that you can write for this?

liamappelbe:
> Is there no end-to-end test that you can write for this?

From the perspective of the native_objc_test, this change makes no difference.

brianquinlan (approved):

PR #332, opened 2022-04-18
Update pubspec.yaml

devoncarew:
- add the `repository` field to the pubspec

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -4,8 +4,8 @@
 
 name: ffigen
 version: 5.0.0-dev.1
-homepage: https://github.com/dart-lang/ffigen
 description: Generator for FFI bindings, using LibClang to parse C header files.
+repository: https://github.com/dart-lang/ffigen
 
 environment:
   sdk: '>=2.16.0-124.0.dev <3.0.0'


pubspec.yaml
```diff
@@ -4,8 +4,8 @@
 
 name: ffigen
 version: 5.0.0-dev.1
-homepage: https://github.com/dart-lang/ffigen
```
brianquinlan, line 7: You can't leave this? The docs say that homepage can also be a repository:
https://dart.dev/tools/pub/pubspec

devoncarew: Yeah, but it doesn't add a lot of value to have both if they both point to the same thing. We want to have the `repository` field populated for core dart packages - it'll let us additional automation and tooling here. If this package had a website that was distinct from the repo url, then I'd say keep both. But given that they're the same, users would just see the same link twice in the pub page:

Screen Shot 2022-04-18 at 6 44 36 PM

dcharkes (approved):

pubspec.yaml
dcharkes: That makes sense. 👍

PR #333, opened 2022-04-19
Block support

liamappelbe:
Add support for parsing and generating generating wrappers for Blocks (Objective C's closure objects).

- Currently blocks can only be created from C function pointers.
- When the block is created, we wrap it in a `_ObjCBlock`  struct, copied from https://clang.llvm.org/docs/Block-ABI-Apple.html
- All Blocks, even with different signatures, share the same singleton `_ObjCBlockDesc`.
- So that Blocks can be deduped by their signature, we give them a fake USR that is based on the cacheKeys of their signature
- For exceptionalReturn in trampoline, it was necessary to add `Type.getDefaultValue`, which returns a sensible default value for the type.
- Future work: calling Blocks from Dart (even those defined in ObjC), and converting Dart closures to Blocks.

#299


test/native_objc_test/block_test.dart
```diff
@@ -0,0 +1,73 @@
...
+      // itself is working correctly.
+      final blockTester = BlockTester.makeFromMultiplier(lib, 10);
+      expect(blockTester.call(123), 1230);
+      final intBlock = blockTester.getBlock();
+      final blockTester2 = BlockTester.makeFromBlock(lib, intBlock);
+      expect(blockTester2.call(456), 4560);
+    });
+
+    test('Block from function pointer', () {
+      final block = ObjCBlock.fromFunctionPointer(
```
brianquinlan, line 63: Is this API meant for end-users? Because this seems like a pretty dangerous (in that it is not type safe) and difficult to use.

liamappelbe: `block.pointer` is an escape hatch like `castFrom`, and I'm hoping to eliminate all the cases where it's necessary. The rest of the API is type safe afaik.

There's a bit of a limitation to how friendly we can make the API, because `Pointer.fromFunction` is restricted in how it can be used. There's static analysis that turns the function passed to it into a C function pointer, and that analysis requires that there is no intermediate stage in between that function and the `Pointer.fromFunction` call. In other words, it's always going to be the case that the user has to call Pointer.fromFunction themselves.

But if you have a specific suggestion about how to make it more ergonomic, I'd love to hear it.

dcharkes: Please help me clarify for myself what we're trying to do here. The mental model that I have for C APIs is that ffigen generates a low level API (with lots of `Pointer`s and only top level functions), and that a user manually wraps that in an object-oriented API. The object-oriented API does not expose any `dart:ffi` classes, only normal Dart classes. (See https://www.youtube.com/watch?v=2MMK7YoFgaA for the mental model.)

Now, because Objective C is already object-oriented, we can immediately generate an object-oriented API. `class BlockTester` does not extend or implement a `dart:ffi` type.

When using a low level C binding without the object-oriented API, a user would just use `Pointer.fromFunction(_add100, 999)`.
In a high level object-oriented API where the API owner owns the callback function. The API owner would just inline the `fromFunction` in the implementation behind the API.
However, in a high level object-oriented API where the API user provides arbitrary callbacks this is not possible.

The typical workaround is to not use `fromFunction` on the users' top-level functions, but create a `Map<int, RetType Function(ArgumentTypes)>` in Dart, store the users' closures there, and use `fromFunction` on a function defined within the API. However, this requires an extra integer argument to be passed to the native library, and the native library should pass that integer back in its callback. This can of course be added if the native library is owned by the person making the Dart API. However, if the native library is not owned by the person writing the object-oriented API, this workaround doesn't work.

I don't see a good workaround for the user having to do `fromFunction` themselves here. The users doing `fromFunction` here is relatively type safe because the `Pointer` has type arguments with the native function signature.

FYI: `fromFunction` needs to statically know its target to be able to do AOT compilation.

I don't see a way for `fromFunction` supporting closures unless users specify how many extra trampoline entries need to be generated, and a hard crash if someone wants one more closure:

* https://github.com/dart-lang/sdk/issues/48865

> block.pointer is an escape hatch like castFrom, and I'm hoping to eliminate all the cases where it's necessary. 

I agree, those cases should be eliminated. The `block` is the objectoriented API, and the `block.pointer` leaks the abstraction by showing the underlying `dart:ffi` object. Can we eliminate it in this PR?

mraleph: I have the following suggestion: unlike naked function pointers, blocks _are_ essentially closures. 

You are already using this (because you are storing `target` separately from the `invoke` trampoline).

You can take this one step further and allow wrapping any Dart closure (not just an FFI pointer to a static function) into a block. Instead of storing target of the call as a function pointer, you can put closure into a persistent handle and store that handle into the block object. Then `invoke` just unwraps the handle and trampolines to the target.

If dealing with persistent handles is too wonky (it requires calling methods of DL API through FFI) you can just setup your own array on the side and use indices into that array instead of handles. 

The only complication here is that we would need to start properly disposing blocks allocated on Dart side to avoid resource leaks and we would need to provide `copy` and `dispose` callbacks that do a bit of reference counting on "closure handles/ids".

But other than that this should work.

liamappelbe: Yeah, I'm planning to support converting closures to blocks as a follow up PR. When you say "blocks are essentially closures", are you suggesting that I shouldn't support the function pointer case at all?

test/native_objc_test/block_test.dart
```diff
@@ -0,0 +1,73 @@
...
+    });
+
+    test('BlockTester is working', () {
+      // This doesn't test any Block functionality, just that the BlockTester
+      // itself is working correctly.
+      final blockTester = BlockTester.makeFromMultiplier(lib, 10);
+      expect(blockTester.call(123), 1230);
+      final intBlock = blockTester.getBlock();
+      final blockTester2 = BlockTester.makeFromBlock(lib, intBlock);
+      expect(blockTester2.call(456), 4560);
```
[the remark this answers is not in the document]
liamappelbe: `blockTester2` is an instance of the `BlockTester` class, which is written in ObjC. It's not actually a block. It's just to let me test calling blocks from inside ObjC. `BlockTester.call` calls the wrapped block, to verify that the block it wraps is constructed correctly, from ObjC's perspective. Maybe I can rename the `.call` method to make this distinction clearer

test/native_objc_test/block_test.dart
mraleph: I think it is fine to support both pointers and full blown closures. 

I was mostly reacting to @dcharkes's comment (and associated feature request https://github.com/dart-lang/sdk/issues/48865 ) --- trying to highlight that https://github.com/dart-lang/sdk/issues/48865 is not actually needed for Obj-C blocks.

brianquinlan (approved):
This is a pretty shallow review because I don't have much context. I'm not sure that the non-closure version implementation will be useful. A lot of APIs require some external context to be useful. For example:

https://developer.apple.com/documentation/foundation/nsdictionary/1414570-enumeratekeysandobjectsusingbloc?language=objc

Without some way of capturing the result of the iteration, you can't effectively use this API.

Issue #335, opened 2022-04-21
Coverage number is inaccurate


liamappelbe:
https://coveralls.io/github/dart-lang/ffigen?branch=master

Coverage is misconfigured. If you look at the covered files, it's only counting files in the test/ directory.

If you look at a much older build, it seems to be working: https://coveralls.io/builds/32799117

PR #337, opened 2022-04-21
Fix coverage configuration

liamappelbe:
https://coveralls.io/builds/48472159

The coverage output now shows lib/ rather than test/

Fixes: #335


.github/workflows/test-package.yml
```diff
@@ -52,13 +52,6 @@ jobs:
         run: dart test/setup.dart
       - name: Run VM tests
         run: dart test --platform vm
-      - name: Collect coverage
-        run: ./tool/coverage.sh
-      - name: Upload coverage
-        uses: coverallsapp/github-action@v1.1.2
-        with:
-          github-token: ${{ secrets.GITHUB_TOKEN }}
-          path-to-lcov: lcov.info
```
mannprerak2, line 61: Are we planning to manually upload coverage?

dcharkes (changes requested):
Master:
image

This PR:
image

Shouldn't we cover both folders?

Also, we could possibly invoke the examples in `example` as well.

dcharkes:
Thanks for cleaning this up @liamappelbe !

.github/workflows/test-package.yml
liamappelbe: When I copied this rule to make the `mac-test` rule below, I accidentally copied this collect coverage stuff too, but we should only have one of these rules collecting coverage. So in this PR I had to delete one of them, and I decided to delete the linux coverage collector, because the mac one also collects coverage for the ObjC tests.

liamappelbe:
> Shouldn't we cover both folders?

The tests themselves aren't usually included in coverage. I'd expect them to have nearly 100% coverage, since they only exist to be run, so it'd artificially inflate the number.

> Also, we could possibly invoke the examples in `example` as well.

I think the tests in `test/example_tests` already do this.

dcharkes (approved):

PR #352, opened 2022-05-03
Fix ObjC @class declarations

### LINKED ISSUE #345: ObjC: NSData methods are properties are not generated
When given the header file for NSData, ffigen produces

```
class NSData extends _ObjCWrapper {
  NSData._(ffi.Pointer<ObjCObject> id, NativeLibrary lib) : super._(id, lib);

  static NSData castFrom<T extends _ObjCWrapper>(T other) {
    return NSData._(other._id, other._lib);
  }

  static NSData castFromPointer(
      ffi.Pointer<ObjCObject> other, NativeLibrary lib) {
    return NSData._(other, lib);
  }
}
```

So most interesting methods are missing.

liamappelbe:
`@class` declarations are ObjC's equivalent of forward declarations. They appear in the AST as an empty instance declaration with no children other than an ObjCClassRef node. So we just need to skip `fillObjCInterfaceMethodsIfNeeded` for these nodes.

Fixes #345


brianquinlan (approved):
Nice!

PR #363, opened 2022-05-11
Remove ObjC tests that check golden generated code

### LINKED ISSUE #362: Remove goldens for platform dependent tests
Our ObjC tests that verify that the output matches a golden can only be regenerated on mac devices. This makes it [hard to contribute](https://github.com/dart-lang/ffigen/issues/7#issuecomment-1124271285) from non-mac devices. These tests aren't super necessary anymore either, since we have good integration test coverage of our ObjC features, in native_objc_test. So we should probably just remove them.

I'm not sure that all the C features are fully covered by integration tests, so we should keep those goldens. I'm only proposing removing the ObjC ones, because those are the ones that are platform dependent.

liamappelbe:
These goldens can only be regenerated on a mac machine, so they make it difficult for people who don't own a mac to contribute. They're also a bit of a nuisance to merge, and we now have good enough coverage with our native_objc_tests that they're not really necessary anymore. The bindings are now generated by native_objc_test/setup.dart, at the same time as the dylibs, and are gitignored.

The objc header_parser tests were removed:
- objc_interface_test was redundant
- objc_basic_types has been merged into native_objc_test/native_objc_test
- objc_config_test was mostly redundant, except for renaming, which has a dedicated test now, called native_objc_test/rename_test

Fixes #362


brianquinlan (approved):
Maybe @dcharkes will have an opinion about this?

dcharkes:
As mentioned in the issue. I think having goldens in general serves a good purpose (and makes reviewing PRs easier when single features are added/tweaked). However, we also want people without Macs to be able to contribute, so removing the goldens for ObjectiveC seems good to me.

dcharkes (approved):

PR #365, opened 2022-05-12
Use ABI-specific integers from `dart:ffi`

simolus3:
Move imports from `package:ffi` back to `dart:ffi`, bump version to 5.0.0.


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 5.0.0
+
+- Stable release targeting Dart 2.17, supporting ABI-specific integer types.s
```
dcharkes, line 3: typo at the end of the sentence.

mannprerak2, line 3: leading `s`

dcharkes (commented):
Can we remove the dependency on `package:ffi` completely?

mannprerak2 (approved):
LGTM

simolus3:
> Can we remove the dependency on `package:ffi` completely?

The ObjC bindings seem to use `calloc` from that package. I'll try to make this package not emit unused imports.

dcharkes:
Sounds good to me.

`calloc` is available in 1.1.2 as well, so we can depend on that instead.

lib/src/header_parser/parser.dart
```diff
@@ -70,8 +70,9 @@ List<Binding> parseToBindings() {
   }
 
   _logger.fine('CompilerOpts used: $compilerOpts');
-  clangCmdArgs = createDynamicStringArray(compilerOpts);
-  final cmdLen = compilerOpts.length;
+  clangCmdArgs = createDynamicStringArray(
+      [...compilerOpts, '-I', '/usr/lib/clang/13.0.1/include/']);
```
dcharkes: Can this path be found automatically?

For example with https://github.com/dart-lang/ffigen/blob/master/lib/src/config_provider/path_finder.dart

Also, I believe the headers that we use are in https://github.com/dart-lang/ffigen/tree/master/third_party/libclang/include/clang-c

dcharkes (approved):
Thanks for the PR Simon!

lib/src/header_parser/parser.dart
dcharkes: Oh, you already beat me to it! 😄

dcharkes:
Thanks @simolus3 ! 🚀

PR #372, opened 2022-05-13
Automatic ref counting for ObjC objects

### LINKED ISSUE #305: ObjC memory management
First pass: expose the manual reference counting methods.
Later: use Dart finalizers to automatically call these methods.

### LINKED ISSUE #351: ObjC: Provide memory management methods
i.e.
```
- retain
- release
- retainCount
```

liamappelbe:
The basic idea is to add `retain` and `release` flags to the constructor of our Dart wrapper objects for ObjC. If `retain` is true, it calls `objc_retain`. If `release` is true, it registers the wrapper object with a `NativeFinalizer` that calls `objc_release` when the wrapper is GC'd.

The wrapper object is constructed at the return site of methods that return an ObjC object, so at that point we have to decide whether to retain and/or release the object. The `retain` flag should be enabled when method doesn't automatically increment the ref count of the returned object (ie, when the user doesn't own the returned reference). So the only time we don't manually call retain is when we already own the returned reference (eg due to `new` or `alloc` calls), which means that our wrapper object always owns the reference it holds. Therefore, the `release` flag should pretty much always be enabled (disabling it creates a weak reference, and we don't really have a use case for that atm).

This probably doesn't handle all the edge cases correctly, but it's a good enough start. One major missing piece is there are annotations that affect this. But we can handle those in follow up CLs as needed.

Fixes #351
Fixes #305


test/native_objc_test/arc_test.dart
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) 2022, the Dart project authors. Please see the AUTHORS file
```
dcharkes, line 1: nit: automated_ref_count_test.dart ?

lib/src/code_generator/objc_built_in_functions.dart
```diff
@@ -144,11 +166,19 @@ class ObjCBuiltInFunctions {
...
     final objType = PointerType(objCObjectType).getCType(w);
     s.write('''
-class _ObjCWrapper {
+class _ObjCWrapper implements ${w.ffiLibraryPrefix}.Finalizable {
   final $objType _id;
   final ${w.className} _lib;
 
-  _ObjCWrapper._(this._id, this._lib);
+  _ObjCWrapper._(this._id, this._lib,
+      {bool retain = false, bool release = false}) {
```
dcharkes, line 174: We have 4 options now here:

1. Do nothing, which is basically do manual memory management.
2. `retain = false`, `release = true`. This would likely be a bug? What is the use case for this?
3. `retain = true`, `release = false`. This is another form of manual memory management, possibly better than 1.
4. Full memory management. In which case the users should take care to never leak `_id`. Is `_id` fully private or can it escape the object. (Which could then result in use after free.)

If 2 is never useful, should we make an enum `MemoryManagement` with 3 cases instead?

It could be useful to have the default value be option 4 instead of option 1.

dcharkes (commented):
High level questions around retain and release:

* Where should the memory management decision be made?
  1. Should it be the API user selecting the memory management for every call?
  2. Or should it be the API generator selecting the memory management for every definition?
  3. Or should it be the API generator selecting a memory management strategy for the whole API?
  * Maybe the 2nd one, because the person generating the bindings double checks if the method already increases the refcount

@brianquinlan what is your experience with the Cocoa API? Which one makes the most sense here?

* And then, if it is option 2, should it be a regex match in the yaml?

* Should it be an enum?
  * One case (the weak references) is likely a bug, leading to use after free.
  * It would be a good place to document the different strategies and when to use them, as you now did on the PR description.

lib/src/code_generator/objc_built_in_functions.dart
```diff
@@ -144,11 +166,19 @@ class ObjCBuiltInFunctions {
 
     final objType = PointerType(objCObjectType).getCType(w);
     s.write('''
-class _ObjCWrapper {
+class _ObjCWrapper implements ${w.ffiLibraryPrefix}.Finalizable {
```
dcharkes, line 169: I think we also want a `release` method that 

1. sets `_isReleased` to true
2. cancels the finalizer
3. decrements the ref-count

And then throw on any method invocations if `_isReleased` is true.

(This only applies for `release = true` constructor calls, so we should also save `_release` in the object.)

liamappelbe:
> High level questions around retain and release:
> 
> * Where should the memory management decision be made?

My plan was to make this decision automatically, per method. That's why I didn't bother with an enum or making the flags user friendly, because it's a private constructor that the user shouldn't be using. If we want to expose this stuff to the user then I'll work on making it nicer.

@brianquinlan Does this approach make sense, or are there cases in the Cacao API where we can't make this decision automatically?

brianquinlan:
> High level questions around retain and release:
> 
> * Where should the memory management decision be made?
>   
>   1. Should it be the API user selecting the memory management for every call?
>   2. Or should it be the API generator selecting the memory management for every definition?
>   3. Or should it be the API generator selecting a memory management strategy for the whole API?
>   
>   
>   * Maybe the 2nd one, because the person generating the bindings double checks if the method already increases the refcount
> 
> @brianquinlan what is your experience with the Cocoa API? Which one makes the most sense here?

I think that (3) isn't going to be viable because one API will have different call types that require different strategies. (2) would be ideal but I think that achieving it will be difficult in practice.

I think if we end up doing with (1) then I'd prefer to just manage my own `release`/`retain` calls - that way I can avoid pointless work e.g. when casting an object temporarily, there is no need to do a `retain`/`release` call.

> 
> * And then, if it is option 2, should it be a regex match in the yaml?
> * Should it be an enum?
>   
>   * One case (the weak references) is likely a bug, leading to use after free.
>   * It would be a good place to document the different strategies and when to use them, as you now did on the PR description.


> > High level questions around retain and release:
> > 
> > * Where should the memory management decision be made?
> 
> My plan was to make this decision automatically, per method. That's why I didn't bother with an enum or making the flags user friendly, because it's a private constructor that the user shouldn't be using. If we want to expose this stuff to the user then I'll work on making it nicer.
> 
> @brianquinlan Does this approach make sense, or are there cases in the Cacao API where we can't make this decision automatically?

I don't know. You could take a look at other objective-c bindings e.g. [PyObjC](https://github.com/ronaldoussoren/pyobjc) and see how they handle these cases.

For example, they do seem to handle the `RETURN_RETAINED` case that I suggested that you add tests for:
https://github.com/ronaldoussoren/pyobjc/blob/master/pyobjc-framework-Quartz/metadata/QuartzComposer.fwinfo#L17

lib/src/code_generator/objc_built_in_functions.dart
liamappelbe: Done

test/native_objc_test/arc_test.dart
liamappelbe: Done

lib/src/code_generator/objc_built_in_functions.dart
liamappelbe: Option 2 is actually what we use for all the objects returned by constructors. It's for the cases where the function returns a reference you own, and you're in charge of releasing it.

dcharkes: Ah, of course, you mentioned this earlier.

dcharkes:
>> My plan was to make this decision automatically, per method.
>
> (2) would be ideal but I think that achieving it will be difficult in practice.

Should we first figure out if this is even possible then?

Maybe this calls for an investigation/design doc analyzing the some ObjectiveC APIs. @mannprerak2 and I did the same for various libraries when needing to make decisions for generating the C bindings. (Some of them made it to this repo as integration tests.)

brianquinlan:
Would it make sense to have an ARC and a manual mode - as clang offers?

Manual mode would expose `release`/`retain` while ARC would generate those calls automatically? And you could start with manual mode.

liamappelbe:
> Would it make sense to have an ARC and a manual mode - as clang offers?

Not sure. If automatic mode works reliably, then I don't think many users would use manual mode. Maybe we can add it if people request it?

liamappelbe:
> I don't disagree but automatic mode has a lot of cases to consider before it can work reliably.

Yeah. I think the most expedient approach is to land this and start using it on real libraries, and fix those cases as we find them. If it turns out that approach turns out to be impractical because there's too many edge cases, then we can add a manual mode.

liamappelbe:
> For a ARC implementation to be minimally viable, you are going to have to support at least all of the cases that I've already identified (except possibly weak references).

Understood. Do you want me to do that now, or can I do it in a follow up CL?

brianquinlan:
> > For a ARC implementation to be minimally viable, you are going to have to support at least all of the cases that I've already identified (except possibly weak references).
> 
> Understood. Do you want me to do that now, or can I do it in a follow up CL?

I don't have a strong opinion about this...but you should only commit this if you think that is moving you in the right direction - otherwise it will just make a mess that you'll have to clean up later.

brianquinlan (commented):
Oops, some of my previous comments didn't appear.

liamappelbe:
> I don't have a strong opinion about this...but you should only commit this if you think that is moving you in the right direction - otherwise it will just make a mess that you'll have to clean up later.

I think it's compatible with those cases. Handling those attributes should just be a matter of passing different flags to the constructor.

test/native_objc_test/arc_test.m
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) 2022, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
```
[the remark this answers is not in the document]
liamappelbe: I have tests for basic "owned" and "borrowed" cases. The other cases aren't supported yet, so I'll add tests for them as I add support those cases.

I removed getTotalObjects and gave each test its own counter though.

test/native_objc_test/automated_ref_count_test.dart
```diff
@@ -0,0 +1,122 @@
...
+      final obj1 = ArcTestObject.alloc(lib).initWithCounter_(counter);
+      expect(counter.value, 1);
+      final obj2 = ArcTestObject.alloc(lib).initWithCounter_(counter);
+      expect(counter.value, 2);
+      final obj3 = ArcTestObject.alloc(lib).initWithCounter_(counter);
+      expect(counter.value, 3);
+
+      // GC to clean up temporaries created between alloc and initWithCounter_.
+      doGC();
+
```
brianquinlan, line 74: Is it worth having a expect(counter.value, 3) here?

brianquinlan (changes requested):
My intent here is please consider my comments and then feel free to merge.

brianquinlan (approved):

Issue #379, opened 2022-05-24
Handle memory management attributes


liamappelbe:
From Brian's comment's [here](https://github.com/dart-lang/ffigen/pull/372#discussion_r874031339).

- [ ] Properties with copy/retain attribute
- [ ] NS_RETURNS_RETAINED
- [ ] Autorelease pools
- [ ] Weak references
- [ ] Delegates

PR #380, opened 2022-05-24
Handle NS_RETURNS_RETAINED and methods that start with new/alloc or contain copy

liamappelbe:
Rather than just exact matching "new" and "alloc" methods.

#379

### DIFF
--- a/test/native_objc_test/automated_ref_count_test.dart
+++ b/test/native_objc_test/automated_ref_count_test.dart
@@ -42,12 +42,20 @@ void main() {
     }
 
     verifyRefCountsInner(Pointer<Int32> counter) {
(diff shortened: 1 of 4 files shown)


test/native_objc_test/automated_ref_count_test.dart
```diff
@@ -42,12 +42,20 @@ void main() {
     }
 
     verifyRefCountsInner(Pointer<Int32> counter) {
-      final obj1 = ArcTestObject.alloc(lib).initWithCounter_(counter);
+      final obj1 = ArcTestObject.new1(lib);
```
brianquinlan, line 45: Is it possible to structure the tests so that it is more clear what scenario you are testing?

brianquinlan (approved):
Nice. The tests are much easier to follow!

PR #388, opened 2022-05-27
Add the ability to control the reference count behavior in castFromPointer

brianquinlan:
Fixes https://github.com/dart-lang/ffigen/issues/387


liamappelbe (commented):
LGTM. What's the use case though? Why not just always have the new object retain and release the reference?

liamappelbe (approved):

PR #391, opened 2022-05-28
Removed config `dart-bool` and target Bool / bool

### LINKED ISSUE #315: Target `Bool` / `bool` instead of wrapping as integers
`dart:ffi` has support for booleans since https://dart-review.googlesource.com/c/sdk/+/216900

We should generate that instead of `!= 0` with integers.

Context: https://github.com/dart-lang/ffigen/pull/311#discussion_r848653161

cc @mannprerak2, seems I forgot about adding this to `dart:ffi` 😄

mannprerak2:
Closes #315 

Booleans are now always generated with `bool` and `ffi.Bool` as it's Dart and C Type respectively.

I am not sure if this is worth upgrading the version to `6.0.0`, but it's technically a breaking change for people using `dart-bool` config.

Here are some options I could think of -
- Bump version to 6.0.0
- Maybe wait for more things and do this PR before actually going for 6.0.0
- Bump version to `5.1.0` and throw an error if `dart-bool` is passed as `false`.

It's possible we could support this config up until 6.0.0, but it required making quite a few code changes which I feel aren't really worth adding just to be removed later.


dcharkes:
> Bump version to 6.0.0

I think that's fine. There's no reason to artificially try to keep the major version number low. And this way we keep the contract that we don't break people if they don't update the major version. 👍

dcharkes (approved):
lgtm! Thanks @mannprerak2 🚀

PR #405, opened 2022-06-19
Bump `package:ffi`

dcharkes:
* https://github.com/dart-lang/ffi/issues/148

We probably never hit it in FFIgen, but better to updated package:ffi.


dcharkes:
Thanks @liamappelbe !

PR #412, opened 2022-06-29
Skip methods that have incomplete types

### LINKED ISSUE #317: ObjC: Imprecise return type generated
This example comes from `NSURLSession.h`:

```
@interface NSURLSessionTask : NSObject <NSCopying, NSProgressReporting>

@property (nullable, readonly, copy) NSURLRequest  *originalRequest;  
```

The generated Dart code is:
```dart
  static ffi.Pointer<ObjCSel>? _sel_originalRequest;
  NSObject get originalRequest {
    _sel_originalRequest ??= _registerName(_lib, "originalRequest");
    final _ret = _lib._objc_msgSend_126(_id, _sel_originalRequest!);
    return NSObject._(_ret, _lib);
  }
```

Note how the return type is `NSObject` instead of `NSURLRequest`.

### LINKED ISSUE #328: ObjC multiple return type warning
There are some methods triggering the multiple return type warning from in the interface parser. Need to investigate to figure out what this means.

```
[WARNING]: Method "changeWithObject:type:index:" in instance "NSOrderedCollectionChange" has multiple return types.
[WARNING]: Method "changeWithObject:type:index:associatedIndex:" in instance "NSOrderedCollectionChange" has multiple return types.
[WARNING]: Method "differenceByTransformingChangesWithBlock:" in instance "NSOrderedCollectionDifference" has multiple return types.
[WARNING]: Method "registeredTypeIdentifiersWithFileOptions:" in instance "NSItemProvider" has multiple return types.
```

### LINKED ISSUE #341: ObjC: Incorrect parameter types generated for some Objective-C methods
For example, the declaration:
```
+ (nullable instancetype)URLWithString:(NSString *)URLString;
```

Results in this generated code:
```
  static NSURL URLWithString(NativeLibrary _lib, NSObject? URLString) {
```

When the type of `URLString` should be `NSString`.

FYI: `castFrom` can be used as a work-around.

### LINKED ISSUE #373: ObjC: Overly general type generated for some methods/properties
This declaration:
```
/*!
    @abstract Returns the HTTP request method of the receiver.
    @result the HTTP request method of the receiver.
*/
@property (nullable, readonly, copy) NSString *HTTPMethod;
```

Generates this code:
```dart
  /// !
  /// @abstract Returns the HTTP request method of the receiver.
  /// @result the HTTP request method of the receiver.
  NSObject? get HTTPMethod {
    final _ret = _lib._objc_msgSend_1(_id, _lib._sel_HTTPMethod1);
    return _ret.address == 0 ? null : NSObject._(_ret, _lib);
  }
```

Notice how the return type is `NSObject?` instead of `NSString?`

### LINKED ISSUE #409: NSCoder.objectZone parse error
[NSCoder.objectZone](https://developer.apple.com/documentation/foundation/nscoder/1442570-objectzone?language=objc) is supposed to return `NSZone*`, but it's being parsed as returning `NSZone`.

### LINKED ISSUE #410: Skip methods that take/return incomplete structs
This is a check that parseFunctionDeclaration already does for C functions. We need to do the same for methods, because taking/returning an incomplete struct causes compile errors when the user tries to use the generated bindings.

liamappelbe:
Fixes #410

Also change how method return types are parsed to fix #409, since I couldn't write the test otherwise. This also fixes #328, fixes #373, fixes #341, and fixes #317

### DIFF
--- a/lib/src/code_generator/objc_interface.dart
+++ b/lib/src/code_generator/objc_interface.dart
@@ -125,7 +125,7 @@ class $name extends ${superType?.name ?? '_ObjCWrapper'} {
     for (final m in methods.values) {
       final methodName = m._getDartMethodName(uniqueNamer);
       final isStatic = m.isClass;
-      final returnType = m.returnType!;
+      final returnType = m.returnType;
 
       // The method declaration.
       if (m.dartDoc != null) {
@@ -397,7 +397,7 @@ class ObjCMethod {
   final String? dartDoc;
   final String originalName;
   final ObjCProperty? property;
-  Type? returnType;
+  final Type returnType;
   final bool isNullableReturn;
   final List<ObjCMethodParam> params;
   final ObjCMethodKind kind;
@@ -412,7 +412,7 @@ class ObjCMethod {
     this.dartDoc,
     required this.kind,
     required this.isClass,
-    this.returnType,
+    required this.returnType,
     this.isNullableReturn = false,
     List<ObjCMethodParam>? params_,
   }) : params = params_ ?? [];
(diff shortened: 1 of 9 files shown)


liamappelbe:
Seems to be making some method return types to generic. I'll have to investigate a bit more.

dcharkes:
> Foundation/NSObject.h

That was the same issue as I had earlier when I tried to re-order finding `libclang.dylib` on MacOS. Maybe something on the MacOS machines in the Github workflow changed?

Though, I could repro it locally (when trying to use libclang from XCode instead of one of the others).

liamappelbe:
> I'm missing some test cases for the bugs that are closed. Should we add those cases to a test to make sure we don't regress them in the future?

They're covered by bad_method_test. They're also covered by the objc example I'm writing. Once I get that working reliably, I think it would make sense to turn that into a test. It'd be good to have some integration tests that are using real mac APIs.

PR #415, opened 2022-07-06
Check that property types aren't incomplete

### LINKED ISSUE #411: NSDecimal error
NSDecimal is a bitfield struct, which is not supported, so it is treated as an incomplete struct. However, this check happens after methods/functions are parsed, so the [usual check](https://github.com/dart-lang/ffigen/issues/410) that ignores functions that take/return incomplete structs doesn't work.

The best fix would probably just be to move the bitfield check earlier in the parsing.

liamappelbe:
You can't use an incomplete struct as a property (it won't compile), but since bit fields aren't supported in FFI, we treat them as incomplete structs. So you can hit this code path by using a bit field as a property.

Fixes #411


dcharkes (approved):
lgtm

I've filed

* https://github.com/dart-lang/ffigen/issues/418

to track by-value support, which is blocked on `dart:ffi` not supporting it.

PR #416, opened 2022-07-06
ObjC example

### LINKED ISSUE #408: Add an Objective C example

liamappelbe:
Add an example that uses the AVFAudio framework to play audio files.

Fixes #408


example/objective_c/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+```
+dart play_audio.dart test.mp3
+```
+
+## Config notes
+
+The ffigen config for an Objective C library looks very similar to a C library.
+The most important difference is that you must set `language: objc`. If you want
+to filter which interfaces are included you can use the `objc-interfaces:`
```
dcharkes, line 14: `objc-interfaces:`? The top level readme.md specifies `objcInterfaces`. We should fix the inconsistency.

example/objective_c/README.md
```diff
@@ -0,0 +1,36 @@
...
+option. This works similarly to the other filtering options.
+
+It is recommended that you filter out just about everything you're not
+interested in binding (see the ffigen config in [pubspec.yaml](./pubspec.yaml)).
+Virtually all Objective C libraries depend on Apple's internal libraries, which
+are huge. Filtering can reduce the generated bindings from millions of lines to
```
dcharkes, line 20: Haha 😄

dcharkes, line 20: Question:

Does objectiveC respect the `include-directives:`? e.g. can users also opt to allow-list some of Apple library header files but skip the rest? (It's probably still cleaner to allow-list the individual interfaces, but `include-directives:` should be an option.)

example/objective_c/pubspec.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+environment:
+  sdk: '>=2.17.0 <3.0.0'
+
+dependencies:
+  ffi: ^2.0.1
+dev_dependencies:
+  ffigen:
+    path: '../../'
+  lints: ^1.0.1
```
dcharkes, line 15: 2.0.0?

example/objective_c/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Generating bindings
+
+At the root of this example (`example/objective_c`), run:
+
+```
+dart run ffigen
+```
+
+This will generate [avf_audio.dart](./avf_audio.dart).
```
dcharkes, line 36: nit: `*_bindings.dart` or `*_generated_bindings.dart`.

example/objective_c/README.md
```diff
@@ -0,0 +1,36 @@
...
+The ffigen config for an Objective C library looks very similar to a C library.
+The most important difference is that you must set `language: objc`. If you want
+to filter which interfaces are included you can use the `objc-interfaces:`
+option. This works similarly to the other filtering options.
+
+It is recommended that you filter out just about everything you're not
+interested in binding (see the ffigen config in [pubspec.yaml](./pubspec.yaml)).
+Virtually all Objective C libraries depend on Apple's internal libraries, which
+are huge. Filtering can reduce the generated bindings from millions of lines to
+tens of thousands.
```
dcharkes, line 21: The generated file is still 60000 LOC. This is not really a problem, but should we consider being able to filter members inside objective c interfaces?

I've filed an issue to track it.

* https://github.com/dart-lang/ffigen/issues/419

dcharkes (approved):
How cool is this! 😎 

Please also add it as test in https://github.com/dart-lang/ffigen/tree/objc_example/test/example_tests so that we run it locally when doing `dart test` and it's run on the CI.

example/objective_c/README.md
liamappelbe: Yeah, every header, C or ObjC, goes through `shouldIncludeRootCursor`, which checks all that stuff.

example/objective_c/README.md
liamappelbe: Yeah, I have a note about that somewhere on my todo list.

dcharkes: Feel free to land it with the 60000 LOC.

(Fun fact: the 60000 LOC file is bigger than the mp3. 😅)

liamappelbe:
The example test is failing because the details of the output are going to depend on the version of the libraries that the github bot has. So instead of doing an exact string comparison, I'm going to just grep for each of the classes and methods I'm using and make sure they exist.

Issue #428, opened 2022-07-27
Delete Dart Functions associated with ObjC Blocks


liamappelbe:
This is not feasible to do at the moment, because [`NativeFinalizers`](https://api.dart.dev/stable/2.17.0/dart-ffi/NativeFinalizer/NativeFinalizer.html) can't call functions created by `Pointer.fromFunction`. See https://github.com/dart-lang/ffigen/issues/381#issuecomment-1196093581 for details.

This will be possible once https://github.com/dart-lang/sdk/issues/47778 is fixed

liamappelbe:
Actually, this could probably also be solved by async callbacks. The Block struct's dispose field doesn't have the restriction mentioned above (that NativeFinalizers can't call functions created by Pointer.fromFunction), because it isn't directly called by the NativeFinalizer (it's called automatically when the Block is cleaned up). We still have the issue that the dispose function can be called on a random thread, but that's what async callbacks are for. This works as long as async callbacks can fail gracefully in the case where the VM is in the middle of shutting down (it's fine to leak the function in that case).

mkustermann:
> This is not feasible to do at the moment, because [NativeFinalizers](https://api.dart.dev/stable/2.17.0/dart-ffi/NativeFinalizer/NativeFinalizer.html) can't call functions created by Pointer.fromFunction. See https://github.com/dart-lang/ffigen/issues/381#issuecomment-1196093581 for details.
> 
> Actually, this could probably also be solved by async callbacks. 

It is feasible by generating C code as outlined in https://github.com/dart-lang/ffigen/issues/381#issuecomment-1196093581. The asynchronous callbacks will inevitably also have to be based on generated C code, for very similar reasons as here (namely that there's objects in the ObjC block that need to be pulled out and used before executing dart code)

One could avoid generating C code by using isolate independent code instead (which is C written in Dart DSL), but current plan is to not have isolate independent code and instead make it seamless for packages to have accompanying native code.

One could also avoid generating C code by baking ObjC specific knowledge into the Dart compiler, but that doesn't seem like a good idea (the Dart compiler would need to rely on assumptions made by a unrelated code generator)

liamappelbe:
> The asynchronous callbacks will inevitably also have to be based on generated C code, for very similar reasons as here (namely that there's objects in the ObjC block that need to be pulled out and used before executing dart code)

[...]

PR #429, opened 2022-07-27
Memory management for Blocks

### LINKED ISSUE #381: Memory management for blocks
Right now we leak all the block pointers.

liamappelbe:
Add a copy/release finalizer for blocks, same as we have for NSObjects. Automatically cleans up the native memory associated with Blocks.

This implements all but the last dot point from https://github.com/dart-lang/ffigen/issues/381#issuecomment-1194456759. Note that this still leaks the Dart Function objects (see https://github.com/dart-lang/ffigen/issues/381#issuecomment-1196093581)

Fixes #381. Filed a separate issue (#428) for cleaning up the Dart Functions, since that's not going to happen any time soon.

### DIFF
--- a/lib/src/code_generator/objc_block.dart
+++ b/lib/src/code_generator/objc_block.dart
@@ -28,6 +28,8 @@ class ObjCBlock extends BindingType {
   BindingString toBindingString(Writer w) {
     final s = StringBuffer();
 
+    builtInFunctions.ensureBlockUtilsExist(w, s);
+
     final params = <Parameter>[];
     for (int i = 0; i < argTypes.length; ++i) {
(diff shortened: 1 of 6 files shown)


brianquinlan (commented):
Would it make sense to just let blocks leak if it is a hack to clean them up?

liamappelbe:
> Would it make sense to just let blocks leak if it is a hack to clean them up?

The way I'm cleaning up blocks in this PR is not a hack. The only hack is in the test that checks the refcount.

The Dart `Function` objects associated with `ObjCBlock.fromFunction` are still leaked, because I wasn't able to clean them up, with or without hacks.

brianquinlan:
Sorry, I misunderstood then.

brianquinlan (approved):

PR #431, opened 2022-07-28
Add exclude-all-by-default config flag

### LINKED ISSUE #422: Maybe add an option to the config to exclude everything by default
We default to including everything, unless the user specifies an include/exclude rule. But for Objective-C we pretty much want to exclude everything by default, then just include a few specific interfaces. So I end up writing configs like this:

```yaml
functions:
  exclude:
    - '.*'
structs:
  exclude:
    - '.*'
unions:
  exclude:
    - '.*'
globals:
  exclude:
    - '.*'
macros:
  exclude:
    - '.*'
enums:
  exclude:
    - '.*'
unnamed-enums:
  exclude:
    - '.*'
objc-interfaces:
  include:
    - 'AVAudioPlayer'
```

It'd be nicer to be able to specify a config flag that excludes `'.*'` by default.

```yaml
exclude-everything-by-default: true
objc-interfaces:
  include:
    - 'AVAudioPlayer'
```

liamappelbe:
Fixes #422

### DIFF
--- a/example/objective_c/README.md
+++ b/example/objective_c/README.md
@@ -18,7 +18,15 @@ It is recommended that you filter out just about everything you're not
 interested in binding (see the ffigen config in [pubspec.yaml](./pubspec.yaml)).
 Virtually all Objective C libraries depend on Apple's internal libraries, which
 are huge. Filtering can reduce the generated bindings from millions of lines to
-tens of thousands.
+tens of thousands. You can use the `exclude-all-by-default` flag, or exclude
+individual sets of declarations like this:
+
+```yaml
+functions:
+  exclude:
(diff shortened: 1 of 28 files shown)


example/objective_c/README.md
```diff
@@ -1,36 +1,44 @@
-# Objective C example
```
liamappelbe, line 1: Recommend you hide whitespace changes for this file. Most of these diffs are just my editor switching the line endings.

dcharkes (approved):
lgtm with comment

Should we add some unit tests testing the config setting, besides only the end-to-end tests with yaml configs?

PR #397, opened 2022-06-07
Fix `include-directives` on Windows

dcharkes:
Closes: https://github.com/dart-lang/native/issues/239

PR #433, opened 2022-08-02
Run Flutter FFI plugin template ffigen on CI

dcharkes:
```
  # Sanity check the latest `flutter create --template plugin_ffi`.
  # This will break if we change the Flutter template or the generated code.
  # But, getting libclang on the LUCI infrastructure has proven to be
  # non-trivial. See discussion on
  # https://github.com/flutter/flutter/issues/105513.
  # If we need to change the generated code, we should temporarily disable this
  # test, or temporarily disable the requirement for all bots to be green to
  # merge PRs.
  # Running this sanity check on one OS should be sufficient. Chosing Windows
  # because it is the most likely to break.
```

Integration test for:

* https://github.com/flutter/flutter/issues/105513

### DIFF
--- a/test_flutter/flutter_template_tests/flutter_plugin_ffi_test.dart
+++ b/test_flutter/flutter_template_tests/flutter_plugin_ffi_test.dart
@@ -1,0 +1,175 @@
+// Copyright (c) 2022, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:test/test.dart';
+
+void main() {
+  Uri? tempDirUri;
+  final projectName = 'test_project';
+
+  setUp(() async {
+    tempDirUri = (await Directory.current.createTemp('.temp_test_')).uri;
+  });
+
+  tearDown(() async {
+    final dir = Directory(tempDirUri!.toFilePath());
+    if (await dir.exists()) {
+      await dir.delete(recursive: true);
+    }
+  });
+
+  test('Run Flutter', () async {
+    final projectDirUri = tempDirUri!.resolve('$projectName/');
+    final libDirUri = projectDirUri.resolve('lib/');
+    final bindingsGeneratedUri =
+        libDirUri.resolve('${projectName}_bindings_generated.dart');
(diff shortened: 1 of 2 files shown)


dcharkes:
The test currently fails because

* https://github.com/dart-lang/ffigen/pull/397

is not in Flutter yet.

dcharkes:
@liamappelbe PTAL

test_flutter/flutter_template_tests/flutter_plugin_ffi_test.dart
```diff
@@ -0,0 +1,178 @@
...
+      print(message);
+      if (throwOnFailure) {
+        throw Exception(message);
+      }
+    }
+    print('Command `$commandString` done.');
+  }
+}
+
+class EnsureExists implements Task {
```
liamappelbe, line 143: This is unused

test_flutter/flutter_template_tests/flutter_plugin_ffi_test.dart
```diff
@@ -0,0 +1,178 @@
...
+
+  test('Run Flutter', () async {
+    final projectDirUri = tempDirUri!.resolve('$projectName/');
+    final libDirUri = projectDirUri.resolve('lib/');
+    final bindingsGeneratedUri =
+        libDirUri.resolve('${projectName}_bindings_generated.dart');
+    final bindingsGeneratedCopyUri =
+        libDirUri.resolve('${projectName}_bindings_generated_copy.dart');
+
+    await Task.serial([
```
liamappelbe, line 33: All this Task stuff seems like a lot of boilerplate. Why not just write a `runProcess` and `copyFile` function, and do:

```dart
await runProcess(...);
await copyFile(...);
await runProcess(...);
```

liamappelbe (approved):
Approved with comments.

test_flutter/flutter_template_tests/flutter_plugin_ffi_test.dart
dcharkes: I had this laying around in another project. I can simplify it for this test if you prefer.

liamappelbe: Feel free to merge as-is if you're too busy to clean it up. It's only a test after all. I've already approved.

Issue #401, opened 2022-06-09
Support `@FfiNative`


dcharkes:
We would like to support generating [`FfiNative` annotations](https://api.dart.dev/stable/2.17.3/dart-ffi/FfiNative-class.html) instead of bindings with based on a `lookup` function.

### Config

The config can be a toplevel switch:

```yaml
ffi-native: true
```

(We could do a filter. But what would be the use case? Conceivably either everything from a header is looked up with the FFI Native Resolver, _or_ with dynamic lookup. Not a mix.)

```yaml
functions:
  ffi-native:
    include:
      # Match function name.
      - 'myFunc'
      # All functions
      - '.*'
    exclude:
      # If you only use exclude, then everything
      # not excluded is `FfiNative`.
      - 'dispose'
```

### Sample output

```c
/** Adds 2 integers. */
int sum(int a, int b);
```

Would generate

```dart
/// Adds 2 integers.
@FfiNative<ffi.Int Function(ffi.Int, ffi.Int>('sum')
external void sum(int a, int b);
```

Note that currently `FfiNative`s are looked up with the lookup function `Dart_SetFfiNativeResolver` in [dart_api.h](https://github.com/dart-lang/sdk/blob/main/runtime/include/dart_api.h). In the future we might want to add support for defining lookups in another way (e.g. `pubspec.yaml`).

This will move `dart:ffi` more in the direction of static binding.

WASM requires static linking, so the [dart2wasm](https://github.com/dart-lang/sdk/tree/main/pkg/dart2wasm) would be using this for FFI to to C files compile to WASM.

#### Top level definitions?

Should the `FfiNative`s be top level or nested inside a class? We do not have a `lookup` function or `DynamicLibrary` to look them up, and the `Dart_SetFfiNativeResolver` can only define a single lookup. So the argument that one could use the bindings for two dynamic librarys / two sets of functions does not apply. So I think it makes sense to have top level definitions.

cc @mannprerak2 Any other design questions that come to mind?

cc @askeksa-google @osa1

mannprerak2:
It'd certainly be simpler for us to keep them inside a class (as we do currently).

mannprerak2:
From a code generation point of view. 

From the user API point of view I feel maybe we should keep all the functions grouped together, since the ffi-native filters might not match all the functions.

dcharkes:
I've been working on `FfiNative`s to support more modes of operation.

[...]

PR #447, opened 2022-08-17
`FfiNative` support

dcharkes:
Closes: https://github.com/dart-lang/ffigen/issues/401


dcharkes:
@mannprerak2 

```yaml
ffi-native:
```

seems to be ignored. The whole `ffiNativeValidator` is not run on it. Is this a known issue?

```yaml
ffi-native:
  asset: # empty
```

```yaml
ffi-native:
  asset: 'asset'
```

both work.

dcharkes:
@mannprerak2 I'm not entirely sure if this is the right way to implement it. Please feel free to let me know a better way to do it.

lib/src/code_generator/writer.dart
```diff
@@ -233,6 +235,12 @@ class Writer {
       s.write('}\n\n');
     }
 
+    if (ffiNativeBindings.isNotEmpty) {
```
liamappelbe, line 238: This if seems redundant.

liamappelbe (approved):
This is cool. Way less boilerplate for function bindings.

example/simple/generated_bindings.dart
```diff
@@ -86,21 +86,21 @@ class NativeLibrary {
       _dividePtr.asFunction<ffi.Pointer<ffi.Float> Function(int, int)>();
 
   /// Divides 2 floats, returns a pointer to double.
-  ffi.Pointer<ffi.Double> dividePercision(
+  ffi.Pointer<ffi.Double> dividePrecision(
     ffi.Pointer<ffi.Float> a,
     ffi.Pointer<ffi.Float> b,
   ) {
-    return _dividePercision(
```
mannprerak2, line 93: 🙈

lib/src/code_generator/func.dart
```diff
@@ -119,21 +118,53 @@ class Func extends LookUpBinding {
...
+      s.write('return $funcVarName');
+
+      s.write('(\n');
+      for (final p in functionType.parameters) {
+        s.write('    ${p.name},\n');
+      }
+      s.write('  );\n');
+      s.write('}\n');
+
+      if (exposeSymbolAddress) {
```
mannprerak2, line 152: Since exposeSymbolAddress is ignored incase of ffi native functions, maybe we should raise a warning log to users with both of these keys in their config.

lib/src/header_parser/sub_parsers/functiondecl_parser.dart
```diff
@@ -80,6 +80,7 @@ Func? parseFunctionDeclaration(clang_types.CXCursor cursor) {
       exposeFunctionTypedefs:
           config.exposeFunctionTypedefs.shouldInclude(funcName),
       isLeaf: config.leafFunctions.shouldInclude(funcName),
+      ffiNativeConfig: config.ffiNativeConfig,
```
mannprerak2, line 83: Is it a possibility that we may have some functions which are ffi native and some which are looked up via a dynamic library later?

If ffinative is a global config(applied to all functions regardless) then maybe we can just keep this boolean at Library level (instead of each Func)

mannprerak2 (approved):

mannprerak2:
> ffi-native:
seems to be ignored. The whole ffiNativeValidator is not run on it. Is this a known issue?

Not known, probably since this is the first key-only config. Will have to investigate here.

lib/src/header_parser/sub_parsers/functiondecl_parser.dart
dcharkes: That's what I did at first, but then the `lookUpBindings` can contain `Func` and `Global`. And `Global`s are not supported.
Maybe it should error on `Global`s when `FfiNative`s are wanted.

The `Func.toBindingString()` has no access to the containing `Library`, so it doesn't know the config of that library. Should we pass it in as an argument to that function? Or should we keep `Binding` fully self contained (meaning easier to unit test). Any preference?

mannprerak2: > Maybe it should error on Globals when FfiNatives are wanted.

It sounds better, we can also just skip generating globals and log a warning (Not supported with ffiNative)

> The Func.toBindingString() has no access to the containing Library

We can propagate ffiNative from library to Writer which is available.

dcharkes: > We can propagate ffiNative from library to Writer which is available.

That kind of feels like the wrong thing, the writer just knows about writing, not about config. If `Func` contains everything that it needs for `toBindingString` that keeps the information more local. So I'll keep it in `Func`.

> It sounds better, we can also just skip generating globals and log a warning (Not supported with ffiNative)

Done.

dcharkes:
> > ffi-native:
> > seems to be ignored. The whole ffiNativeValidator is not run on it. Is this a known issue?
> 
> Not known, probably since this is the first key-only config. Will have to investigate here.

Fixed.

dcharkes:
Thanks for the quick reviews @mannprerak2 and @liamappelbe! 🚀

PR #455, opened 2022-09-15
Work around analyzer bug for FfiNatives

dcharkes:
Work around https://github.com/dart-lang/sdk/issues/49684 to turn CI green again


dcharkes:
@liamappelbe please land this to turn the CI green again.

I'd like to find some time to fix the analyzer itself, but it will keep erroring on Dart stable until we release a Dart 2.19, so it's better to just avoid the issue in the tests for now.

PR #456, opened 2022-09-22
Remove list_diff package

liamappelbe:


TzviPM:
Was this not helpful in the end? Or just not worth the dependency? Should a different implementation be used, then? Or are we fine with just removing this entirely?

liamappelbe:
> Was this not helpful in the end? Or just not worth the dependency? Should a different implementation be used, then? Or are we fine with just removing this entirely?

It was a convenience that made a couple of bugs easier to diagnose. If I need something like this again I'll just add this change back to my local repo, or roll my own diff. If it's going to make it more annoying to import into g3 then it's not worth it.

dcharkes:
> Was this not helpful in the end? Or just not worth the dependency? Should a different implementation be used, then? Or are we fine with just removing this entirely?

We're fine with removing it entirely. It is only used for making failing tests have a nicer output. But I tend to use a diff between two files for that anyway instead of relying on the output of the test.

dcharkes (approved):
Thanks for the quick PR @liamappelbe!

PR #462, opened 2022-09-27
Remove ignoreFilter from typedef parsing.

### LINKED ISSUE #460: Config typedef include/exclude no longer works.
It seems that the config `typedefs -> exclude/include` is no longer working (Probably since 6.1.0).
I tracked this change down to 
https://github.com/dart-lang/ffigen/blob/e262c9a1c6d32d69c8434be98e466e504a6a7f7a/lib/src/header_parser/sub_parsers/typedefdecl_parser.dart#L33-L42

First added in (6.1.0) https://github.com/dart-lang/ffigen/pull/431/files#diff-aae6aaed3ca43f5c68f72ffbb689be5b81821be3b37f925deb09ada6b1412697R42

Is this intentional? Simply removing `ignoreFilter` from here should otherwise solve this for C bindings.

@liamappelbe @dcharkes

mannprerak2:
Closes #460

- Bumped version to `7.0.0`
- Added/Updated tests


dcharkes (commented):
We need a test case so that it doesn't break again in the future.

dcharkes:
@liamappelbe could you please take a look at how to make it work again for the C bindings without breaking ObjectiveC?

liamappelbe:
The fix ended up being a bit more involved than I thought. I've pushed some changes to this branch (I think? haven't tried do do this before)

I had to special case the 'instancetype' typedef always be allowed. This fixed most of the tests, but the block tests were still broken.

The block API has changed a bit, because the old typedef behavior caused raw block pointers to be returned in some cases where the Dart wrapper should have been returned (I think there was a typedef that was adding a layer of indirection that confused my block special casing). The new API is actually more ergonomic, so that's a good change.

I also had to add an `ObjCBlock.fromBlockPointer` function as an escape hatch for the block test, because it's now expecting the wrapper rather than the raw pointer.

liamappelbe:
@mannprerak2 I'll look into the test/native_objc_test/block_test.dart failure

mannprerak2:
@liamappelbe could you also look into `test/large_integration_tests/large_test.dart: large_test Libclang test` it seems to be failing probably because of a different llvm version on macos-latest (See https://github.com/dart-lang/ffigen/actions/runs/3138470319/jobs/5097866684)
Maybe we should alter the source header - `third_party/libclang/include/clang-c/Index.h` since this test will fail depending on the llvm version used.

liamappelbe:
Sounds like another occurrence of https://github.com/dart-lang/ffigen/issues/432

liamappelbe:
`ObjCBlock.fromBlockPointer` was the wrong fix. Switched `getBlockRetainCount` to take a void* instead.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# 6.2.0
```
dcharkes, line 1: Take care with the merging order.

dcharkes (approved):

dcharkes:
Thanks @mannprerak2 and @liamappelbe! 🚀

liamappelbe:
Re-enabled swift_class_test, because #467 is caused by typedefs. Since this PR stops a lot of unnecessary typedefs from generating, it causes that test to pass again. It doesn't actually fix the underlying bug though, so I'll leave the issue open.

mannprerak2:
@liamappelbe Seems like we updated the version to `7.0.0-dev` on master, but we should probably have released a `6.2.0` first.

liamappelbe:
Why not just put this PR in v7? We can release v7 whenever.

mannprerak2:
@dcharkes we can merge this now. I've bumped the version to `7.0.0`.

dcharkes:
Thanks @mannprerak2 and @liamappelbe !

PR #463, opened 2022-09-28
Use `dart_style` over `dart` CLI for formatting

TzviPM:
Using `dart_style` directly instead of invoking `dart format` avoids an extra layer of indirection.

### DIFF
--- a/lib/src/code_generator/library.dart
+++ b/lib/src/code_generator/library.dart
@@ -4,11 +4,10 @@
 
 import 'dart:io';
 
-import 'package:cli_util/cli_util.dart';
+import 'package:dart_style/dart_style.dart';
 import 'package:ffigen/src/code_generator.dart';
 import 'package:ffigen/src/config_provider/config_types.dart';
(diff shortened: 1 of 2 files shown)


dcharkes:
If I remember correctly, importing the package and running it from source was significantly slower than using `dart format`. Since `dart format` is a precompiled utility it does not need to JIT compile the whole package.

Could you elaborate on your use-case?

TzviPM:
> If I remember correctly, importing the package and running it from source was significantly slower than using `dart format`. Since `dart format` is a precompiled utility it does not need to JIT compile the whole package.
> 
> Could you elaborate on your use-case?

That makes sense. I didn't realize this was an explicit design choice. In my case, I am invoking ffigen as part of a sandboxed build process in which invoking the `dart` CLI would have to be explicitly whitelisted for `ffigen`. As it turns out, I can just as easily disable formatting, since the generated files will be consumed only programatically by later steps in the build process.

PR #465, opened 2022-09-29
Assume relative paths are relative to the config file

### LINKED ISSUE #443: All paths should be relative to config directory
All the relative paths in the config file should be treated as relative to the config directory. This would allow running ffigen from any directory.

This is blocking adding a test for the swift example.

liamappelbe:
Adds swift_example_test.dart, now that it's unblocked.

Fixes #443


lib/src/config_provider/spec_utils.dart
```diff
@@ -366,36 +374,6 @@ bool headersValidator(List<String> name, dynamic value) {
   }
 }
 
-String libclangDylibExtractor(dynamic value) => getDylibPath(value as String);
```
liamappelbe, line 369: Dead code

dcharkes (approved):
lgtm

Thanks @liamappelbe !

PR #464, opened 2022-09-28
Use tmpDir over working directory

TzviPM:
Fixes #400.

PR #466, opened 2022-09-29
Fix mac bot

liamappelbe:
Mac bot is broken at head: https://github.com/dart-lang/ffigen/actions/runs/3136139841

This is blocking several PRs: #462 #464 #465

The failures are caused by the mac bot updating to Apple clang version 13.0.0 (clang-1300.0.29.30). One of the failures is a recurrence of #432, where some generated comments are different on different versions of clang, so I made that test ignore that diff. The other failure is #467, where swift_class_test.dart is legitimately broken on this version of clang. I'll need to dig into this failure, so I've disabled the test for now.


lib/src/header_parser/clang_bindings/clang_bindings.dart
```diff
@@ -1176,6 +1176,17 @@ class Clang {
   late final _clang_getCursorKindSpelling =
       _clang_getCursorKindSpellingPtr.asFunction<CXString Function(int)>();
 
+  /// Return a version string, suitable for showing to a user, but not
+  /// intended to be parsed (the format is not guaranteed to be stable).
+  CXString clang_getClangVersion() {
```
liamappelbe, line 1181: I used this for debugging, but it seems like a handy function to keep around.

PR #468, opened 2022-10-04
Symbol file import export

### LINKED ISSUE #22: Sharing shared definitions
Consider the following scenario:

* header1 - with a struct
* header2 which imports header1 - having that struct in a signature
* header3 which imports header1 - having also that struct in the signature

Now we run the generator on header 2 and 3: they both generate the same struct.
These two structs will be Dart classes with the same name but incompatible.
You can work around that with `Pointer.cast<>()`, but that's not safe.
Also, this increases code size.

It would be better to be able to run the generator on all 3 headers files and let the generated files for header 2 and 3 import the generated file for header 1.

We need to think about a good design for this, especially if headers 2 and 3 are in different dart packages.

(I am not sure how common this scenario is. For example libclang and sqlite3 stand very much on themselves. But I can imagine structs defined in OS APIs fall into this category, so it would be nice to address this.)

mannprerak2:
Closes #22 

### Summary -
A library can now export the symbols by specifying
```yaml
output:
  ...
  symbol-file: 
    output: '../generated/base_symbols.yaml' # Warns if not using a package: URI.
    import-path: '../generated/base_gen.dart' # Warns if not using a package: URI.
```

Exported `symbol.yaml` file looks like
```yaml
format_version: "1.0.0"
files:
  ../generated/base_gen.dart:
    used-config:
      ffi-native:false
    symbols:
      c:@E@BaseEnum:
        name: "BaseEnum"
      c:@F@base_func1:
        name: "base_func1"
```

To import symbols,- 
```yaml
import:
  symbol-files:
    - '../generated/base_symbols.yaml' # Package Uri is also allowed here
```

Doing so will
1. Type map the shared types (Structs, Unions, Typedefs) from symbol-file.
2. Other types with same USRs (such as functions, enums) will not be generated (since they are already defined in the imported library)


mannprerak2:
@dcharkes @mahesh-hegde Just a small POC as of now, checkout - [`example/shared_bindings`](https://github.com/mannprerak2/ffigen/tree/symbol_file_import_export/example/shared_bindings).

dcharkes (commented):
Nice!

Some high level questions. (Maybe better discussed on https://x.com/dart-lang/ffigen/issues/22 rather than the PR.)

example/shared_bindings/ffigen_configs/a_shared_base.yaml
```diff
@@ -0,0 +1,16 @@
...
+name: NativeLibraryASharedB
+description: Bindings to `headers/a.h` with shared definitions from `headers/base.h`.
+output: '../generated/a_shared_b_gen.dart'
+headers:
+  entry-points:
+    - '../headers/a.h'
+library-imports:
+  base_h: '../generated/base_gen.dart'
+symbol-file-map:
+  base_h: '../generated/base_symbols.yaml'
```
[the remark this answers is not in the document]
mannprerak2: Sounds good.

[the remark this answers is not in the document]
mannprerak2: Implemented this.

I also went ahead to remove the `symbol-file-map` config completely and merging it with `library-imports`.
So the config now looks like this - 
```yaml
library-imports:
  old_format: 'old/format/still/works.dart'
  base_h:
    import-path: '../generated/base_gen.dart'
    symbol-file: '../generated/base_symbols.yaml'
```

I.m.o we should keep `import-path` mandatory, since to make it optional, we need to take import-path as input from the user while generating symbols.yaml file as well.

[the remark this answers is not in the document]
mannprerak2: I guess that's true.

The problem I was trying to avoid solving was of invalid import-path for shared bindings (in a single project) if a user specifed a relative path.
We can either only allow package uri path for import-path in symbols.dart or allow the end user to override the import-path instead.
wdyt?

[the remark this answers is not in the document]
dcharkes: > The problem I was trying to avoid solving was of invalid import-path for shared bindings (in a single project) if a user specifed a relative path.

That could indeed be a problem.
If the user does intend to make it available to other packages, their first release will reveal the issue.
On the other hand, if the user is not intending to make the bindings available to other packages, it could be fine to have a relative path (though only if you only import it from specific locations in your project, so probably still not what you want).

> We can either only allow package uri path for import-path in symbols.dart or allow the end user to override the import-path instead.

Maybe we should just add a warning when generating.

mannprerak2:
@dcharkes I've added a summary in the top comment (https://github.com/dart-lang/ffigen/pull/468#issue-1396892992).

Let me know when you think this is feature ready, I can begin to add tests, update readme, etc.

dcharkes:
> @dcharkes I've added a summary in the top comment ([#468 (comment)](https://github.com/dart-lang/ffigen/pull/468#issue-1396892992)).

Thanks! That's a very good write up 👍 
I've added my reply in the issue https://github.com/dart-lang/ffigen/issues/22, then it'll be easier to find for me later.

lib/src/code_generator/utils.dart
```diff
@@ -70,3 +70,73 @@ String makeDoc(String text) {
 
   return s.toString();
 }
+
+class YamlWriter {
```
dcharkes, line 74: package:yaml_edit can write yaml files.

```dart
String _toYamlString(Object yamlEncoding) => (YamlEditor('')
      ..update(
          [],
          wrapAsYamlNode(
            yamlEncoding,
            collectionStyle: CollectionStyle.BLOCK,
            scalarStyle: ScalarStyle.DOUBLE_QUOTED,
          )))
    .toString();
```

If this file becomes to burdensome to maintain, we could depend on that package. (The Dart SDK itself already depends on it internally.)

example/shared_bindings/generated/base_symbols.yaml
```diff
@@ -0,0 +1,26 @@
+format_version: "1.0.0"
```
dcharkes, line 1: beauty!

example/shared_bindings/generated/base_symbols.yaml
```diff
@@ -0,0 +1,26 @@
...
+files:
+  ../generated/base_gen.dart:
+    used-config:
+      ffi-native:false
+    symbols:
+      c:@E@BaseEnum:
```
dcharkes, line 7: are all of the characters ever occurring in here valid yaml keys? Or should we have quotes around them?

dcharkes (approved):
lgtm with comments (and api discussion in issue)

lib/src/code_generator/utils.dart
mannprerak2: Oh cool, I've replaced my small implementation with it.

example/shared_bindings/generated/base_symbols.yaml
mannprerak2: package:yaml is able to parse these, so I think these are okay.
Anyway, since I've replaced my implementation with package:yaml_edit's it should take care of generating a valid syntax I suppose.

mannprerak2:
@dcharkes And 3 weeks later, this is finally ready 😝 . I've added tests, updated the readme, and bumped the version.

mannprerak2:
@dcharkes it seems like macros can have a different USR after all. Not sure why this is the case, but the windows test are failing due to this.
Here's a macro USR- `c:base.h@496@macro@BASE_MACRO_1`. The number (496 here) can possibly change

For now I've simply removed macros from example/shared_bindings.

dcharkes:
> @dcharkes And 3 weeks later, this is finally ready 😝 . I've added tests, updated the readme, and bumped the version.

No problemo! Thanks for all your contributions to this open source project!! 🚀 

> it seems like macros can have a different USR after all. Not sure why this is the case, but the windows test are failing due to this. Here's a macro USR- `c:base.h@496@macro@BASE_MACRO_1`. The number (496 here) can possibly change
> 
> For now I've simply removed macros from example/shared_bindings to prevent tests from failing on windows.

Should we refuse to emit macros in the symbols file then? And emit an error? (And a test that checks that an error is emitted.)

mannprerak2:
> Should we refuse to emit macros in the symbols file then?

Yeah I guess we can, alt

> And emit an error? 

Do you mean we raise an error and exit without generating the symbol file? Or is it a just warning which mentions that macros have been removed since they cannot be cross-referenced reliably?

---

I think we're probably seeing this since C allows the redefinition of macros, (albeit with a warning from the compiler). 
```red
/headers/a.h:17:9: warning: 'BASE_MACRO_1' macro redefined [Lexical or Preprocessor Issue]
```

example/shared_bindings/README.md
```diff
@@ -0,0 +1,10 @@
+# Shared bindings
+
+An example to showcase how bindings can share types
+with other bindings
+
+## Generating bindings
+At the root of this example (`example/shared_bindings`), run -
+```
+./run.sh
```
dcharkes, line 9: You could consider making this a Dart script (with `Process.run`s) instead of a shell script.

example/shared_bindings/ffigen_configs/a_shared_base.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+name: NativeLibraryASharedB
+description: Bindings to `headers/a.h` with shared definitions from `headers/base.h`.
+output: '../lib/generated/a_shared_b_gen.dart'
+headers:
+  entry-points:
+    - '../headers/a.h'
+import:
+  symbol-files:
+    # Both package Uri and file paths are allowed here.
```
dcharkes, line 13: allowed -> supported

example/shared_bindings/ffigen_configs/base.yaml
```diff
@@ -0,0 +1,17 @@
+# Copyright (c) 2022, the Dart project authors. Please see the AUTHORS file
+# for details. All rights reserved. Use of this source code is governed by a
+# BSD-style license that can be found in the LICENSE file.
+
+name: NativeLibraryBase
+description: Bindings to `headers/base.h`.
+output:
+  bindings: '../lib/generated/base_gen.dart'
+  symbol-file:
+    # Although file paths are allowed here, prefer Package Uri's here
```
dcharkes, line 10: allowed -> supported

dcharkes: period at end of sentence.

example/shared_bindings/lib/generated/base_symbols.yaml
```diff
@@ -0,0 +1,22 @@
...
+      c:@S@BaseStruct2:
+        name: BaseStruct2
+      c:@U@BaseUnion1:
+        name: BaseUnion1
+      c:@U@BaseUnion2:
+        name: BaseUnion2
+      c:base.h@T@BaseTypedef1:
+        name: BaseTypedef1
+      c:base.h@T@BaseTypedef2:
+        name: BaseTypedef2
```
dcharkes, line 22: `package:yaml_edit` does not put a newline at the end of the file? Or do we `trim()`?

test/test_utils.dart
```diff
@@ -55,11 +55,46 @@ String _normalizeGeneratedCode(
...
+    library: library,
+    pathForActual: pathForActual,
+    pathToExpected: pathToExpected,
+    fileWriter: ({required Library library, required File file}) =>
+        library.generateFile(file),
+    codeNormalizer: codeNormalizer,
+  );
+}
+
+/// Generates actual file using library and tests using [expect] with expected
```
dcharkes, line 68: period at end of sentence.

lib/src/strings.dart
```diff
@@ -131,6 +135,27 @@ const sizemap_native_mapping = <String, int>{
 // Library imports.
 const libraryImports = 'library-imports';
 
+// Sub Keys of symbol file.
+const symbols = 'symbols';
+
+// Symbol file yaml.
+const formatVersion = "format_version";
+
+/// Current symbol file format version. This is generated when generating any
```
dcharkes, line 144: https://dart.dev/guides/language/effective-dart/documentation#do-start-doc-comments-with-a-single-sentence-summary

```dart
/// Current symbol file format version.
///
/// This is generated when generating any ...
```

dcharkes (approved):
Sweet stuff!

lib/src/strings.dart
mannprerak2: Ah, how did I miss this.

example/shared_bindings/lib/generated/base_symbols.yaml
mannprerak2: It's yaml_edit, I've made sure to append a newline if the yaml_edit doesn't.

example/shared_bindings/README.md
mannprerak2: Done

PR #469, opened 2022-10-05
Fix some noisy warnings

### LINKED ISSUE #306: Unsupported pack value -2
We get a lot of this error when running the ObjC generator over internal headers (eg, this comes up in test/native_objc_test/native_objc_test.dart):

```
WARNING : Unsupported pack value "-2" for Struct "__CFString".
```

This isn't specific to `__CFString`, it happens for a lot of ObjC structs.

### LINKED ISSUE #307: Invalid enum constant
We get a lot of this error when running the ObjC generator over internal headers (eg, this comes up in test/native_objc_test/native_objc_test.dart):

```
SEVERE  : Invalid enum constant.
```

Step 1 is to improve the log message so that we know which enum is causing the problem.

liamappelbe:
Fix some noisy warnings that often show up in ObjC binding generation.

Fixes #306 
Fixes #307

### DIFF
--- a/lib/src/header_parser/clang_bindings/clang_bindings.dart
+++ b/lib/src/header_parser/clang_bindings/clang_bindings.dart
@@ -2522,6 +2522,32 @@ abstract class CXTypeNullabilityKind {
   static const int CXTypeNullability_Invalid = 3;
 }
 
+/// List the possible error codes for \c clang_Type_getSizeOf,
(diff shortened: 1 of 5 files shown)


dcharkes (approved):
lgtm

We might want tests that test the warnings we output. As of now, this PR is untested. (I worry slightly that the tests might be fragile though if we test on the actual string contents of the warnigs. But writing tests that only check the error types might be non-trivial.)

Issue #470, opened 2022-10-05
Blank bindings file for Objective-C headers via ffigen


leighajarett:
I am trying to use ffigen to generate the dart bindings for some of Apple's APIs. However, every one I have tried (with the exception of the examples in the documentation) have resulted in a blank bindings file. 

I am on a Mac, and I am am using dart version 2.19

Example 1
[Core Motions API](https://developer.apple.com/documentation/coremotion)

- Pubspec 

```
ffigen:
    name: IntentsLibrary
    language: objc
    output: "foundation_bindings.dart"
    exclude-all-by-default: true
    objc-interfaces:
      include:
        - "CMMotionManager"
    headers:
      entry-points:
        - "/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/System/Library/Frameworks/CoreMotion.framework/Headers/CMMotionManager.h"
```

- Output
```
Building package executable... (1.4s)
Built ffigen:ffigen.
Running in Directory: '/Users/leighaj/dart/ffi_example/ffi_example'
[SEVERE] : Objective C support is EXPERIMENTAL. The API may change in a breaking way without notice.
[WARNING]: Prefer adding Key '[description]' to your config.
Input Headers: [/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/System/Library/Frameworks/CoreMotion.framework/Headers/CMMotionManager.h]
[SEVERE] : Header /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/System/Library/Frameworks/CoreMotion.framework/Headers/CMMotionManager.h: Total errors/warnings: 1.
[SEVERE] :     /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/System/Library/Frameworks/CoreMotion.framework/Headers/CMMotionManager.h:77:117: warning: 'visibility' attribute ignored [Semantic Issue]
Finished, Bindings generated in /Users/leighaj/dart/ffi_example/ffi_example/foundation_bindings.dart
```

Example 2
[HealthKit](https://developer.apple.com/documentation/healthkit/hkhealthstore?language=objc)

- Pubspec
```
ffigen:
    name: IntentsLibrary
    language: objc
[...]

PR #471, opened 2022-10-05
Remove unnecessary root cursor filtering

liamappelbe:
This was added early on in objective c binding development, to reduce the amount of code generated. But now that we have interface filtering and exclude-all-by-default it's not necessary (and in some cases it's too aggressive and excludes the interface the user wants).

#470


dcharkes (approved):

dcharkes:
Thanks @liamappelbe !

PR #474, opened 2022-10-11
Add a step for building cursor definition map.

### LINKED ISSUE #179: Empty Opaque structs generated due to forward declaration.
In C, you can make forward declarations like this -
```C
struct A;

struct A
{
    int a;
};
```

ffigen currently generates `struct A` as `Opaque` (without any fields). (This probably also affects enum declarations).

Note: This will only happen when parsing this declaration sequentially (i.e if the declaration was included).

mannprerak2:
Closes #179 
- Handle declarations with definitions accessible from a different entry point header.
- Added tests
- Update version, changelog.

C allows us to forward declarations without any reference as to where it is defined. This can lead to different entry-point header ordering, resulting in some Structs being generated as Opaque (and skipped declarations which depend on these by value since they are incomplete)

This is fixed by adding another step before parsing the declarations. 
1. Parse all the entry-points, and build a map (`CursorIndex`) of USR to CXCursor  (only for Structs/Unions/Enums currently).
2. When parsing these declaration types to binding
    1. We find its definition in current translation unit with libclang (`clang_getCursorDefinition`).
    2. If not found, we use `CursorIndex` for finding the CXCursor definition since it may be in another translation unit.
    3. If not found, log a warning for user that the definition could not be found (So they may add more entry-points if need be)

### DIFF
--- a/test/header_parser_tests/separate_definition_test.dart
+++ b/test/header_parser_tests/separate_definition_test.dart
@@ -1,0 +1,50 @@
+// Copyright (c) 2020, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:ffigen/src/code_generator.dart';
+import 'package:ffigen/src/config_provider.dart';
+import 'package:ffigen/src/header_parser.dart' as parser;
+import 'package:ffigen/src/strings.dart' as strings;
+import 'package:logging/logging.dart';
+import 'package:test/test.dart';
+import 'package:yaml/yaml.dart' as yaml;
+
+import '../test_utils.dart';
+
+late Library actual, expected;
+
+void main() {
+  group('separate_definition', () {
+    setUpAll(() {
+      logWarnings(Level.SEVERE);
+    });
+    test('different header order', () {
+      final entryPoints = [
+        "test/header_parser_tests/separate_definition_base.h",
+        "test/header_parser_tests/separate_definition.h"
+      ];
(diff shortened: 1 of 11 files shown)


test/header_parser_tests/separate_definition_test.dart
```diff
@@ -0,0 +1,50 @@
...
+late Library actual, expected;
+
+void main() {
+  group('separate_definition', () {
+    setUpAll(() {
+      logWarnings(Level.SEVERE);
+    });
+    test('different header order', () {
+      final entryPoints = [
+        "test/header_parser_tests/separate_definition_base.h",
```
dcharkes, line 24: Can we make two tests, also one with the order of the entryPoints reversed? Then we test that it really doesn't matter what the order is.

dcharkes (approved):
Nice! 🚀 

@jonasfj could you verify with a git dependency if this solves your issue from https://github.com/google/webcrypto.dart/pull/41#pullrequestreview-1120953775?

test/header_parser_tests/separate_definition_test.dart
mannprerak2: The current test will generate 2 bindings, one with headers - h1, h2 and one with headers - h2, h1

This should ensure that different header order will generate the same bindings.

What other test should I add?

dcharkes: Oh, I was still very sleepy this morning. You're already doing `entryPoints.reversed`! 😄

mannprerak2: 😆

PR #477, opened 2022-10-19
Reuse Blocks in ObjCBlock.fromFunction

### LINKED ISSUE #476: ObjCBlock.fromFunction should return same block for same function
Calling `ObjCBlock.fromFunction` twice with the same function should return the identical `ObjCBlock`. This would help a bit with #428, by creating fewer persistent handles.

It should be pretty easy to, using an `Expando` to attach the `ObjCBlock` to the Dart function. Then, inside `ObjCBlock.fromFunction`, we just check if the user's function has an `ObjCBlock` attached already, and return it if it's there.

liamappelbe:
When `ObjCBlock.fromFunction` is passed in the same function multiple times, return a new `ObjCBlock` that wraps the same underlying block struct. Explicitly retain (increase ref count) of the block struct in this case. This helps reduce the leak caused by #428 by reducing the number of blocks created if the user accidentally creates many blocks from the same function.

This is accomplished by sticking the block struct pointer in an `Expando` (`_funcToBlock`), with the Dart function as the key.

I also tried sticking the `ObjCBlock` directly in `_funcToBlock`, which was simpler and cleaner, but this caused the `ObjCBlock` to never be GC'd, because the Dart function is reachable via the `_ObjCBlock_closureRegistry`. We can try refactoring this to put the `ObjCBlock` in the `_funcToBlock` once #428 is fixed.

Fixes #476

### DIFF
--- a/lib/src/code_generator/objc_block.dart
+++ b/lib/src/code_generator/objc_block.dart
@@ -103,8 +103,8 @@ $voidPtr $registerClosure(Function fn) {
     final exceptionalReturn = defaultValue == null ? '' : ', $defaultValue';
     s.write('''
 class $name extends _ObjCBlockBase {
-  $name._(${blockPtr.getCType(w)} id, ${w.className} lib) :
-      super._(id, lib, retain: false, release: true);
+  $name._(${blockPtr.getCType(w)} id, ${w.className} lib, {
+      bool retain = false}) : super._(id, lib, retain: retain, release: true);
 
   /// Creates a block from a C function pointer.
   $name.fromFunctionPointer(${w.className} lib, $natFnPtr ptr) :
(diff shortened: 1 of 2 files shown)


lib/src/code_generator/objc_block.dart
```diff
@@ -115,12 +115,20 @@ class $name extends _ObjCBlockBase {
...
+    if (oldBlock != null) {
+      return $name._(oldBlock, lib, retain: true);
+    }
+    final newBlock = lib.${builtInFunctions.newBlock.name}(
           _dartFuncTrampoline ??= ${w.ffiLibraryPrefix}.Pointer.fromFunction<
               ${trampFuncType.getCType(w)}>($closureTrampoline
-                  $exceptionalReturn).cast(), $registerClosure(fn)), lib);
+                  $exceptionalReturn).cast(), $registerClosure(fn));
+    _funcToBlock[fn] = newBlock;
+    return $name._(newBlock, lib);
```
mkustermann, line 128: We don't increase refcount for `newBlock` here. 

What happens if the underlying obj-c block is de-rereferenced and freed while still being in the expando map? If one creates another block for same closure, we have a use-after-free, don't we?

I'll comment more on the github issue

mkustermann: I guess the assumption is that the dart wrapper will keep a refcount as long as it's alive?

liamappelbe: Yeah, the Dart wrapper always holds one reference. In all the other cases where we construct the wrapper function, we're using a block pointer from Block_copy, which already has an incremented ref count.

PR #479, opened 2022-10-24
Prevent warnings in FfiNative example

dcharkes:
`FfiNative` will be deprecated and we will introduce `Native` (with slightly different syntax).

See discussion

* https://github.com/dart-lang/sdk/issues/50097

And CL: https://dart-review.googlesource.com/c/sdk/+/265084

Adding a lint ignore here to prevent this example from starting to throw warnings when this lands.


liamappelbe (approved):

dcharkes:
Thanks @liamappelbe and @mannprerak2 !

PR #487, opened 2022-11-17
Remove the built in flag from ObjC interfaces

### LINKED ISSUE #484: Missing NSString constructor when generating the bindings for healthkit
I am using ffigen to generate the bindings for the HealthKit ObjC framework. 

Here is the pubspec.yaml:

```
ffigen:
  name: HealthKit
  description: "Create bindings for the HealthKit APIs"
  language: objc
  output: "lib/healthkit_bindings.dart"
  compiler-opts:
    - '-F/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/System/Library/Frameworks'
  exclude-all-by-default: true
  objc-interfaces:
    include:
      - "HKHealthStore"
      - "HKQuery"
  headers:
    entry-points:
      - "/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/System/Library/Frameworks/HealthKit.framework/Headers/HKHealthStore.h"
      - "/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/System/Library/Frameworks/HealthKit.framework/Headers/HKQuery.h"
```

In my dart file I am creating the library and attempting to convert a dart string to an NSString:

```
final lib = hk.HealthKit(DynamicLibrary.process());
var localeString = hk.NSString(lib, 'en_US')
```

But I get the error: `The class 'hk.NSString' doesn't have an unnamed constructor.`

Let me know if you need any more info!

liamappelbe:
I was using it to try to make sure we only add helper methods like those on NSString to the official versions of those interfaces. This would prevent a user defined NSString from getting those methods. But the flag is unreliable, so in some cases we end up omitting the helper methods when they should be there. So it's better to just remove this restriction.

Fixes #484

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,14 @@
+# 7.2.1
+
+- Fix helper methods sometimes missing from NSString.
+
 # 7.2.0
+
 - Added support for sharing bindings using `symbol-file` config. (See `README.md`
 and examples/shared_bindings).
 
 # 7.1.0
+
 - Handle declarations with definition accessible from a different entry-point.
 
 # 7.0.0
--- a/lib/src/code_generator/objc_built_in_functions.dart
+++ b/lib/src/code_generator/objc_built_in_functions.dart
@@ -311,7 +311,6 @@ class $name implements ${w.ffiLibraryPrefix}.Finalizable {
         (ObjCInterface(
           originalName: "NSData",
           builtInFunctions: this,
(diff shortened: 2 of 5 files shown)


dcharkes (approved):
Thanks @liamappelbe !

PR #491, opened 2022-11-29
Add common code path for test configs

### LINKED ISSUE #490: Add a common code path for config files used in tests
Many tests have code like the following that's used to generate a config file from a yaml string: https://github.com/dart-lang/ffigen/blob/master/test/collision_tests/decl_type_name_collision_test.dart#L21-L30.

Funnelling these through a single code path would make it easier to experiment with changes to the config spec, as codemods could be applied in a single place during testing.

Also, Google's internal use of ffigen requires changes to the file paths for the header files used in tests, and this would decrease the maintenance burden of those changes

TzviPM:
Fixes #490.


test/example_tests/objective_c_example_test.dart
```diff
@@ -22,9 +19,21 @@ void main() {
     });
 
     test('objective_c', () {
-      final pubspecFile = File('example/objective_c/pubspec.yaml');
-      final pubspecYaml = loadYaml(pubspecFile.readAsStringSync()) as YamlMap;
-      final config = Config.fromYaml(pubspecYaml['ffigen'] as YamlMap);
```
dcharkes, line 27: This means we have the config in two places now, how are we going to keep that sync? These tests are here to ensure our examples keep working.

What is the reason for moving the config here instead of reading it from the file?

test/example_tests/objective_c_example_test.dart
```diff
@@ -22,9 +19,21 @@ void main() {
...
+${strings.output}: 'avf_audio_bindings.dart'
+${strings.excludeAllByDefault}: true
+${strings.objcInterfaces}:
+  ${strings.include}:
+    - 'AVAudioPlayer'
+${strings.headers}:
+  ${strings.entryPoints}:
+    - '/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/AVFAudio.framework/Headers/AVAudioPlayer.h'
+${strings.preamble}: |
+  // ignore_for_file: camel_case_types, non_constant_identifier_names, unused_element, unused_field, return_of_invalid_type, void_checks, annotate_overrides, no_leading_underscores_for_local_identifiers, library_private_types_in_public_api
```
dcharkes, line 35: Nit: prefer ignore_for_file on separate lines and alphabetically ordered.

dcharkes (changes requested):

test/example_tests/objective_c_example_test.dart
TzviPM: Yes, I brought this up with Liam as a potential point of contention offline and we discussed a solution. My proposed solution is to make an additional test helper for reading a config from a file given its path from the project root and to use separate config files instead of a key in pubspec.yaml, so that the config file helper can simply read the file into a string and then use the existing helper, rather than parsing the pubspec.yaml to get the ffigen key

test/example_tests/objective_c_example_test.dart
TzviPM: I believe this was copied verbatim from the existing pubspec.yaml. see my previous comment, though, about how I plan to make this easier to synchronize by using a standalone config.yaml

test/example_tests/objective_c_example_test.dart
dcharkes: sgtm

test/example_tests/objective_c_example_test.dart
dcharkes: Please fix the place where this is copied from while your at it. 🙏 

Great! Thanks @TzviPM !

example/simple/ffigen.yaml
```diff
@@ -0,0 +1,6 @@
+name: NativeLibrary
+description: Bindings to `headers/example.h`.
+output: 'generated_bindings.dart'
+headers:
+  entry-points:
+    - 'headers/example.h'
```
liamappelbe, line 6: nit: missing newline

example/swift/ffigen.yaml
```diff
@@ -0,0 +1,19 @@
...
+    'SwiftClass': 'swift_module'
+headers:
+  entry-points:
+    - 'swift_api.h'
+preamble: |
+  // ignore_for_file: camel_case_types, non_constant_identifier_names
+  // ignore_for_file: unused_element, unused_field, return_of_invalid_type
+  // ignore_for_file: void_checks, annotate_overrides
+  // ignore_for_file: no_leading_underscores_for_local_identifiers
+  // ignore_for_file: library_private_types_in_public_api
```
liamappelbe, line 19: nit: missing newline

liamappelbe (approved):
Thanks for cleaning up!

liamappelbe:
Oh, looks like a few tests are failing

dcharkes:
@TzviPM is this still needed for your work?

example/swift/swift_api.h
```diff
@@ -1,4 +1,4 @@
-// Generated by Apple Swift version 5.5.2 (swiftlang-1300.0.40.106 clang-1300.0.29.21)
+// Generated by Apple Swift version 5.7.2 (swiftlang-5.7.2.135.5 clang-1400.0.29.51)
```
dcharkes, line 1: We haven't pinned clang on the CI?

example/swift/config.yaml
```diff
@@ -0,0 +1,19 @@
...
+    'SwiftClass': 'swift_module'
+headers:
+  entry-points:
+    - 'swift_api.h'
+preamble: |
+  // ignore_for_file: camel_case_types, non_constant_identifier_names
+  // ignore_for_file: unused_element, unused_field, return_of_invalid_type
+  // ignore_for_file: void_checks, annotate_overrides
+  // ignore_for_file: no_leading_underscores_for_local_identifiers
+  // ignore_for_file: library_private_types_in_public_api
```
dcharkes, line 19: nit: newline at end of file

example/simple/config.yaml
```diff
@@ -0,0 +1,6 @@
+name: NativeLibrary
+description: Bindings to `headers/example.h`.
+output: 'generated_bindings.dart'
+headers:
+  entry-points:
+    - 'headers/example.h'
```
dcharkes, line 6: nit: newline at end of file

dcharkes (approved):
lgtm

Issue #493, opened 2022-12-09
Missing iOS file when generating bindings for CMPedometer


leighajarett:
Hi! I'm trying to use generate bindings for the CMPedometer class that's part of the CoreMotion framework. But, I am just getting a blank bindings file.

Config:
```
name: PedometerBindings
description: "Bindings for CM pedometer"
language: objc
output: 'lib/pedometer_bindings_generated.dart'
compiler-opts:
  - '-F/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/System/Library/Frameworks'
objc-interfaces:
  include:
    - "CMPedometer"
headers:
  entry-points:
    - "/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/System/Library/Frameworks/CoreMotion.framework/Headers/CMPedometer.h"

```

There are a series of Nullability Issue warnings, such as:
```
[SEVERE] :     /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/stdlib.h:361:13: warning: pointer is missing a nullability type specifier (_Nonnull, _Nullable, or _Null_unspecified) [Nullability Issue]
```

As well as a Lexical or Preprocessor Issue:
```
[SEVERE] :     /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/System/Library/Frameworks/Foundation.framework/Headers/NSURL.h:10:9: fatal error: 'Foundation/NSURLHandle.h' file not found [Lexical or Preprocessor Issue]
```

Sorry in advance if I am missing something!!

liamappelbe:
The nullability issue is a red herring. The second error is the real problem. One of the headers is trying to include Foundation/NSURLHandle.h and can't find it. This usually means that you're missing an include directory or a framework in your compiler-opts.

leighajarett:
It seems like that file should be in the directory used for compiler-opts?

leighajarett:
Actually I don't see that header file (NSURLHandle.h) in the Foundations framework

brianquinlan:
I think that you just need to add the entry point explicitly. Like:

```
headers:
  entry-points:
[...]

PR #495, opened 2022-12-13
Allow user compiler options to override built-in options

liamappelbe:
If clang sees 2 of the same flag with different values, it takes the second. So if we insert user options second they can override our defaults. In #493, this is needed so that the user can target iOS by overriding the default mac sysroot that the ObjC flags default to.

### DIFF
--- a/lib/src/header_parser/parser.dart
+++ b/lib/src/header_parser/parser.dart
@@ -56,21 +56,24 @@ List<Binding> parseToBindings() {
   final index = clang.clang_createIndex(0, 0);
 
   Pointer<Pointer<Utf8>> clangCmdArgs = nullptr;
-  final compilerOpts = List<String>.from(config.compilerOpts);
+  final compilerOpts = <String>[];
 
-  /// Add compiler opt for comment parsing for clang based on config.
-  if (config.commentType.length != CommentLength.none &&
-      config.commentType.style == CommentStyle.any) {
-    compilerOpts.add(strings.fparseAllComments);
-  }
-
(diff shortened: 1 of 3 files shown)


lib/src/header_parser/parser.dart
```diff
@@ -56,22 +56,25 @@ List<Binding> parseToBindings() {
   final index = clang.clang_createIndex(0, 0);
 
   Pointer<Pointer<Utf8>> clangCmdArgs = nullptr;
-  final compilerOpts = List<String>.from(config.compilerOpts);
+  final compilerOpts = <String>[];
```
dcharkes, line 59: Nit: can we make this a list with inline ifs?

```dart
final compilerOpts = <String>[
  // Add compiler opt for comment parsing for clang based on config.
  if (config.commentType.length != CommentLength.none &&
      config.commentType.style == CommentStyle.any)
    compilerOpts.add(strings.fparseAllComments),
  // If the config targets Objective C, add a compiler opt for it.
  if (config.language == Language.objc) ...[
      ...strings.clangLangObjC,
      ..._findObjectiveCSysroot(),
  ],
  // Add the user options last so they can override any other options.
  ...config.compilerOpts
];
```

dcharkes (approved):
lgtm with comments

dcharkes:
Ah, Github is not as smart as Gerrit, it auto merges even with comments...

dcharkes:
Thanks @liamappelbe !

Addressing my own comment in https://github.com/dart-lang/ffigen/pull/496 before publishing.

PR #515, opened 2023-02-12
Support for generating Variadic functions

### LINKED ISSUE #503: Support variadic functions
We'll be adding support for variadic functions to `dart:ffi`:

* https://github.com/dart-lang/sdk/issues/38578

_One interesting thing to think about is how this would work for bindings generation from `.h` files: It wouldn't know what specific var-args signature to use when generating method(s) for a vararg method. Either it could generate a lookup of pointer and require users to `fp.asFunction` or one would need to specify in the configuration._

_Originally posted by @mkustermann in https://github.com/dart-lang/sdk/issues/38578#issuecomment-1369696674_

Generating the function pointer would always require a cast, because a `Pointer<NativeFunction<X Function(X, VarArgs(X, X))>>` can't have the right signature. So it should probably be a `Pointer<Void>` if we go for that option. Which would result in:

```dart
somePointer.cast<NativeFunction<...>>.asFunction<...>();
```

However, that would not allow us to ever use variadic functions with static linking (`@FfiNative`) through ffigen.

Moreover, that requires the user to both write out the C type and Dart type of both the non-variadic arguments and variadic arguments.

So, I believe it would be better to specify the variadic arguments wanted in the config. That way the user only has to write out the C types, and only for the variadic arguments (not for the arguments before the `...`).

Also, since Dart doesn't have overloading, we would need to invent names if we want a function more than once. Adding a number is a bad idea, because when adding a variadic args config it might renumber, maybe we can default to appending the variadic argument types? Or default to appending the variadic argument types if there's more than one config?

```yaml
functions:
  variadic-arguments:
    'printf':
      - [uint32_t, double] # Single config generates `printf`.
```

```yaml
functions:
  variadic-arguments:
    'printf':
      - [uint32_t, double] # Generates `printfUInt32Double`.
      - [uint32_t, char*]  # Generates `printfUInt32CharPointer`.
```

```yaml
functions:
  variadic-arguments:
    'printf':
      - [uint32_t, double] # Generates `printfUInt32Double`.
      - types: [uint32_t, char*]
         postfix: 'IntString' # Generates `printfIntString`.
```

Related issue:

* https://github.com/dart-lang/ffigen/issues/182

@mannprerak2 @mkustermann @Sunbreak WDYT? Any alternative ideas?

mannprerak2:
Closes #503 

Just a draft. I've added a test but it currently skips formatting.

@dcharkes feel free to take a look and suggest changes, although we should probably wait for Dart 3 to merge this.


dcharkes:
Tracking bug for formatter:

* https://github.com/dart-lang/dart_style/issues/1128

test/header_parser_tests/varargs_test.dart
```diff
@@ -0,0 +1,58 @@
...
+${strings.headers}:
+  ${strings.entryPoints}:
+    - 'test/header_parser_tests/varargs.h'
+
+${strings.functions}:
+  ${strings.varArgFunctions}:
+    myfunc:
+      - [int, char*, SA]
+    myfunc2:
+      - [char*, long**]
```
dcharkes, line 35: Nice!

lib/src/config_provider/spec_utils.dart
```diff
@@ -327,6 +328,97 @@ Map<String, ImportedType> makeImportTypeMapping(
...
 
+Type makePointerToType(Type type, int pointerCount) {
+  for (var i = 0; i < pointerCount; i++) {
+    type = PointerType(type);
+  }
+  return type;
+}
+
+String makePostfixFromRawVarArgType(List<String> rawVarArgType) {
+  return rawVarArgType
```
dcharkes, line 339: Should we consider generating a C file with configurable imports instead? And run clang on it to get the mapping from C to Dart types.

For example: What if I use a struct as vararg that I did not transitively included otherwise?

dcharkes (commented):
Nice! 🚀

lib/src/config_provider/spec_utils.dart
mannprerak2: > What if I use a struct as vararg that I did not transitively included otherwise?

That will generate types with no reference, but the current approach allows also using types declaration in any external library as well (library-imports config). So i guess this is a  tradeoff.

dcharkes: Hm, what about renames? If a user has made a rename from `MyStruct` in C to `PrefixedMyStruct` in Dart?

It feels slightly more robust to make it go through the whole pipeline, we would pick up all the features we have in ffigen. But maybe I am mistaken, what do you think?

mannprerak2: I feel it may be more difficult to trace back what renamed a struct than just looking at the generated bindings.

And while C allows structs/unions/typedefs to share the same name, these get renamed by ffigen due to name collision, although this is rare but still possible. The only way to then differentiate those is by having the user specify what sort of type it is with its original name in C.

dcharkes:
This work can be rebased on the master branch now that that's on 3.0

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 8.0.0-dev.3
+- Added support for variadic functions using config `functions -> variadic-arguments`.
```
dcharkes, line 2: nit space

dcharkes (approved):
lgtm!

Issue #520, opened 2023-02-20
Specify LLVM path through CLI / environment variable


mahesh-hegde:
Currently it is possible to specify LLVM path through YAML file, but sometimes it's necessary to do this on single system, while leaving the default locations working for anyone who clones the code and runs `ffigen` on a standard system.

For example, I have LLVM installed via `scoop` which means it's not on default location. Same can happen with less common linux distributions.

It would be nice to be able to specify this through command line or environment variable. The latter will also work with scripts which import `package:ffigen`.

dcharkes:
Maybe better move the discussion on the issue instead of the PR.

JNIgen has a mechanism for being able to override any yaml config value with a -dfoo.bar=baz CLI args. Should we consider something similar for environment variables? And to have this same overriding logic for FFIgen? @mannprerak2 @mahesh-hegde @HosseinYousefi

mannprerak2:
Sounds good.

- we'll need to handle both override and extend behaviour (E.g for compiler opts the current behaviour is to extend the yaml config with a higher priority to cmd args)
- Can we handle lists in args? and what about list of lists?

mahesh-hegde:
> Should we consider something similar for environment variables?

I think environment variables are useful only when you have machine-specific configuration you don't want checked in as code.

Also in this case, ffigen imported as library cannot straightforwardly change the same in config parsed from a YAML file, becase there's no setter in `ffigen.Config` for `llvmPath`. So env var is more useful.

> And to have this same overriding logic for FFIgen

In JNIgen tests and CI we use `--override` to frequently switch between output configs, bindings mode etc.. (we have some configuration explosion there 😃 ). By far that's most common use in CI as well as during dev. Does ffigen have similar requirements?

> Can we handle lists in args?

In JNIgen, List can be specified (separated by `;` IIRC because windows powershell interprets `,` differently). But I don't think it's used much. Mostly because we never specified absolute paths in config.

mahesh-hegde:
> Why not add the scoop locations to the default locations?

a. Scoop installs to user home directory, so that varies based on system.

b. This issue is not really specific to scoop. Fedora has a diff location too. (seems #221 addressed this through adding llvmPath to YAML config, but if I am temporarily using a fedora system, and don't want to change the pubspec, I am out of luck, which is original reason for this issue as well).

> Or we get 3 places how things can be configured: yaml/CLI/env-variables

Valid point. counterpoints

[...]

PR #521, opened 2023-02-22
Read LLVM path from environment variable.

mahesh-hegde:
Closes: #520 

What is a recommended place to adding test case for this? I think we can't set environment variables from within Dart.


dcharkes:
JNIgen has a mechanism for being able to override any yaml config value with a `-dfoo.bar=baz` CLI args. Should we consider something similar for environment variables? @mannprerak2 @mahesh-hegde

mannprerak2:
Yeah that sounds better. It's not always override though, e.g compiler-opts from cmd line and yaml config are both used.

mahesh-hegde:
Any updates related to this?

PR #527, opened 2023-03-08
Bump dependencies and fix lints

dcharkes:
No manual coding was involved, only application of automatic fixes.


example/shared_bindings/lib/generated/a_shared_b_gen.dart
```diff
@@ -4,7 +4,7 @@
 //
 // Generated by `package:ffigen`.
 import 'dart:ffi' as ffi;
-import 'package:shared_bindings/generated/base_gen.dart' as _imp1;
+import 'package:shared_bindings/generated/base_gen.dart' as imp1;
```
mannprerak2, line 7: Applied fix to a generated file. We should probably just add `no_leading_underscores_for_library_prefixes` to preamble in `a_shared_base.yaml`.

mannprerak2: Ah, I didn't see that you changed `const defaultSymbolFileImportPrefix = 'imp';` this as well. 🙈

liamappelbe (approved):

PR #532, opened 2023-03-10
Automatically generate `ignore_for_file: type=lint` 

### LINKED ISSUE #270: Help people generating ignores. Perhaps generate them by default.
See https://github.com/dart-lang/pana/issues/1011

mannprerak2:
Closes #270 

- Automatically generate `ignore_for_file: type=lint` if not specified in the preamble.


mannprerak2:
@dcharkes I didn't add a config parameter to tweak this behavior, since I couldn't think of any use case where one would want to do that.

dcharkes (approved):
lgtm based on rational in https://github.com/dart-lang/ffigen/issues/270#issuecomment-1437236478

Should we keep some tests without this lint, because we control the set of lints in this repo?
That way we can ensure the generated code is in the style that we like in _this_ package. (Instead of trying to be in the style of something that any user is asking for in their package.)

dcharkes:
> Do you have any examples in mind?

I fixed the underscore import name for example from the recommended lints.

Maybe it's not that important.

Issue #522, opened 2023-02-26
Parameter names in function pointer field types


mahesh-hegde:
context: `jnigen`'s bindings to JVM C APIs are generated by a vendored patch of jnigen. I am trying to move it to use the API or at least import the internal `src/` APIs, in the interest of reducing the vendored code.

By default ffigen doesn't generate parameter names for struct fields of function pointer types, although the field declaration can potentially include these names.

I had a clunky patch which added a separate array of these param names.

https://github.com/dart-lang/jnigen/blob/1d858c058b40d8b06735dfdd59c71afde388392b/jni/third_party/ffigen_patch_jni/lib/src/header_parser/sub_parsers/compounddecl_parser.dart#L231

https://github.com/dart-lang/jnigen/blob/1d858c058b40d8b06735dfdd59c71afde388392b/jni/third_party/ffigen_patch_jni/lib/src/header_parser/sub_parsers/compounddecl_parser.dart#L263

Is there a better implementation strategy than "keep a global list and visit all parameter nodes"? If not I will attempt to upstream this.

dcharkes:
@mannprerak2 is probably most suited to answer that question.

mahesh-hegde:
1. Should this handle nested function pointer declarations?

2. Should this apply to variables with global linkage as well? Particularly, we can't get parameter names from `CXType`, we need to visit the cursor of declaration itself.

Now I think there are 2 choices.

* We can change of `getCodeGenType(CXType)` to `getCodeGenType(cursor)`?

* Alternatively we can write the auxiliary visitor to modify the names of params.

cc: @mannprerak2

mannprerak2:
> Should this handle nested function pointer declarations?

This will probably require some significant code refactoring since we'd need to change `getCodeGenType` to work with CXCursor as well as CXType (perhaps both).

The patch is only for a single level of function pointer declaration (only for struct members in their current state). A complete solution will require that refactoring though.

mahesh-hegde:
Makes sense. I will upstream my patch and make a PR.

mahesh-hegde:
Now I am wondering, whether this param names be actually exposed in `external ffi.Pointer<ffi.NativeFunction<...>>` declaration?

i.e

`external external ffi.Pointer<ffi.NativeFunction<ffi.Int Function(ffi.Int, ffi.Int)>>comparator;`

vs 

[...]

PR #537, opened 2023-03-20
Parameter names in function pointer fields

mahesh-hegde:
closes: #522 

TODO: Decide whether it's worth to include parameter names in field `Function` type, after reviewing updated outputs.

If not including them in output, should also figure out a mechanism to test it without generating output.


test/header_parser_tests/struct_fptr_fields.h
```diff
@@ -0,0 +1,15 @@
...
+{
+    // Function pointer field, but no parameters.
+    int (*func1)(void);
+    // Function pointer field with parameters.
+    int (*comparator)(int a, int b);
+    // Function pointer field with lot of parameters
+    int (*veryManyArguments)(double a, float b, char *c, int d, long long e);
+    // Function pointer field with parameters, but no names
+    int (*argsDontHaveNames)(int, int, int, float, char *);
+};
```
dcharkes, line 15: Nit: missing newline at end of file.

dcharkes (commented):
> TODO: Decide whether it's worth to include parameter names in field Function type, after reviewing updated outputs.

It looks useful to me! 🚀

lib/src/code_generator/func_type.dart
```diff
@@ -54,6 +55,21 @@ class FunctionType extends Type {
       p.type.addDependencies(dependencies);
     }
   }
+
+  void addParameterNames(List<String> names) {
+    if (names.length != parameters.length) {
+      return;
+    }
+    final paramNamer = UniqueNamer({});
```
mahesh-hegde, line 63: What's the correct way to rename parameters here so that none will clash with a dart keyword?

mannprerak2: UniqueNamer will handle all dart keywords defined in the keywords.dart file

lib/src/header_parser/sub_parsers/compounddecl_parser.dart
```diff
@@ -271,7 +271,20 @@ int _compoundMembersVisitor(clang_types.CXCursor cursor,
         if (mt.baseType is UnimplementedType) {
           parsed.unimplementedMemberType = true;
         }
-
+        if (mt is PointerType && mt.child is NativeFunc) {
```
mannprerak2, line 274: It's possible that there can be multiple levels of typedef nesting (for mt or mt.child), You can use Type.baseType and/or Type.typedefType to get the underlying type)

mannprerak2 (changes requested):

lib/src/header_parser/sub_parsers/compounddecl_parser.dart
```diff
@@ -326,6 +339,19 @@ int _compoundMembersVisitor(clang_types.CXCursor cursor,
...
+    clang_types.CXCursor parent, Pointer<Void> clientData) {
+  if (cursor.kind == clang_types.CXCursorKind.CXCursor_ParmDecl) {
+    final spelling = cursor.spelling();
+    if (spelling.isNotEmpty) {
+      _paramNameList.add(spelling);
+    }
+  }
+  return clang_types.CXChildVisitResult.CXChildVisit_Continue;
+}
+
```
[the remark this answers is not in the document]
mahesh-hegde: > Then it would be a one-liner to add parameter names to function pointers in function arguments as well in functiondecl_parser.

I am only cursorily (!) familiar with ffigen codebase. Can you elaborate more?

test/header_parser_tests/struct_fptr_fields.h
```diff
@@ -0,0 +1,16 @@
...
+struct S
+{
+    // Function pointer field, but no parameters.
+    int (*func1)(void);
+    // Function pointer field with parameters.
+    int (*comparator)(int a, int b);
+    // Function pointer field with lot of parameters
+    int (*veryManyArguments)(double a, float b, char *c, int d, long long e);
+    // Function pointer field with parameters, but no names
+    int (*argsDontHaveNames)(int, int, int, float, char *);
```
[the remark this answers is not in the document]
mannprerak2: Sure.
```C
typedef int (*CXCursorVisitor)(int cursor, int parent, int client_data);
typedef int (SomeTypedef)(int a);
struct S
{
    int (*comparator)(int a, int (*func1)(void)); // Nested function pointer
    CXCursorVisitor a; // Function pointer from Typedef
    SomeTypedef *b; // Function pointer from a pointer to Typedef
};

```

lib/src/header_parser/sub_parsers/compounddecl_parser.dart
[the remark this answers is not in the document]
mannprerak2: If we separate this out to another sub_parser, we can call that function here as well
https://github.com/dart-lang/ffigen/blob/7f14959e8cfe799fd31524ce6b9b36256cad8e0b/lib/src/header_parser/sub_parsers/functiondecl_parser.dart#L102-L107

Something like this -
```dart
if (pt is PointerType && pt.child is NativeFunc) {
    parseParametersFromFunctionPointerCursor(pt, paramCursor);
}
```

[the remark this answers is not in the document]
mahesh-hegde: On a side note, should I

```dart
if (pt.baseType is NativeFunc) {
  /// ... 
}
```

because somewhere double pointer to function is used?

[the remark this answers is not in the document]
mannprerak2: Yeah we should probably handle pointer relations, and there could be multiple levels of Typedefs too.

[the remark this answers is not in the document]
mannprerak2: > I think it would make more sense to add a variation of getCodegenType(CXCursor) than thi

Yeah you can do that too.

Well I guess using a stack doesn't seem to be necessary in this case (since parsing the types and adding the parameters is a different step) but I only suggested it since other sub parsers follow a similar structure.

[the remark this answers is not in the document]
mahesh-hegde: > Well I guess using a stack doesn't seem to be necessary

If we came this far, why not handle nested function pointers?

I am in favor of changing `toCodeGenType` even if it takes longer. Because we would then be able to have param names in function pointer typedefs, function pointers as parameters etc..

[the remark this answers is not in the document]
mannprerak2: Yeah that would actually be a complete solution.

mahesh-hegde:
Not getting time for implementing all cases.

If the partial feature is useful in itself, can we add full implementation to the backlog and merge this? Or else feel free to discard the PR, since I can always vendor my changes in jnigen.

Cases not covered:

* typedefs
* function pointer with a function pointer parameter
* Maybe more..

mannprerak2:
> If the partial feature is useful in itself, can we add full implementation to the backlog and merge this?

I guess. Wdyt @dcharkes?

mahesh-hegde:
@mannprerak2 I tried to move the logic into getCodeGenType(), but I have a gut feeling that this implementation might not be very reliable.

Look at the changes in source code & reference bindings, and tell what you think.

lib/src/header_parser/utils.dart
```diff
@@ -249,8 +249,8 @@ String? removeRawCommentMarkups(String? string) {
 
 extension CXTypeExt on clang_types.CXType {
   /// Get code_gen [Type] representation of [clang_types.CXType].
-  Type toCodeGenType() {
-    return getCodeGenType(this);
+  Type toCodeGenType({clang_types.CXCursor? originalCursor}) {
+    return getCodeGenType(this, originalCursor: originalCursor);
```
mannprerak2, line 253: Do we need this? why not just make add an extension method `toCodeGenType()` to CXCursor as well?

lib/src/header_parser/type_extractor/extractor.dart
```diff
@@ -314,8 +324,31 @@ Type _extractFromFunctionProto(clang_types.CXType cxtype) {
...
     parameters: parameters,
     returnType: clang.clang_getResultType(cxtype).toCodeGenType(),
-  ));
+  );
+  _parseAndMergeParamNames(functionType, cursor);
+  return NativeFunc(functionType);
+}
+
+// TODO: Should we add a maxDepth argument here, to prevent deep recursion?
+void _parseAndMergeParamNames(
```
mannprerak2, line 336: Probably, unless we are sure that recursion will always be finite.

dcharkes:
> > If the partial feature is useful in itself, can we add full implementation to the backlog and merge this?
> 
> I guess. Wdyt @dcharkes?

I think a partial feature is useful, we should make sure to file an issue with the cases not covered.
Maybe even add a test case for a case that's not covered and have it reference to the open bug.

> @mannprerak2 I tried to move the logic into getCodeGenType(), but I have a gut feeling that this implementation might not be very reliable.
> 
> Look at the changes in source code & reference bindings, and tell what you think.

> reference bindings

lgtm

> source code

@mannprerak2 can lgtm this part.

dcharkes (approved):
lgtm

lib/src/header_parser/sub_parsers/function_type_param_parser.dart
```diff
@@ -0,0 +1,54 @@
...
+import 'dart:ffi';
+
+import '../clang_bindings/clang_bindings.dart';
+import '../data.dart';
+import '../utils.dart';
+
+/// This type holds the list of `ParmDecl` nodes of a function type declaration.
+class FunctionTypeParams {
+  final List<String> paramNames;
+  final Map<String, CXCursor> params;
```
mannprerak2, line 14: Default implementation of Map in Dart is ordered, so you could technically skip the extra array if you like.

mannprerak2 (approved):
LGTM.

lib/src/header_parser/sub_parsers/function_type_param_parser.dart
mahesh-hegde: Is this a guarantee we can rely on? I don't have a preference here.

mahesh-hegde: Well separating concept of names and cursors seems clearer to me. Let's leave it as it is.

dcharkes: Yep, maps are ordered (there is a separate data store and index, the index stores the entry-number in the data store). A "traditional" hashmap stores a pointer the object directly in the index and has no separate data store. 😄

dcharkes:
@mahesh-hegde we missed the pubspec.yaml version bump. 🙃

PR #552, opened 2023-04-13
Explicit '' as name for anon structs/unions

TzviPM:
The C11 spec states that anonymous unions / structs imply they are unnamed:

> An unnamed member whose type specifier is a structure specifier with no tag is called an anonymous structure; an unnamed member whose type specifier is a union specifier with no tag is called an anonymous union.

(See https://port70.net/~nsz/c/c11/n1570.pdf)

That said, in some environments, I noticed the `nested_parsing_test` failing with a union in `Struct6` that should be called `unnamed` instead called `union (anonymous at /.../ffigen/test/header_parser_tests/nested_parsing.h:47:5)`

This CL explicitly uses an empty string for the name of an anonymous union or struct

### DIFF
--- a/lib/src/header_parser/sub_parsers/compounddecl_parser.dart
+++ b/lib/src/header_parser/sub_parsers/compounddecl_parser.dart
@@ -300,16 +300,21 @@ int _compoundMembersVisitor(clang_types.CXCursor cursor,
         // otherwise they will be added in the next iteration.
         if (!cursor.isAnonymousRecordDecl()) break;
 
+        // Anonymous members are always unnamed. To avoid environment-
+        // dependent naming issues with the generated code, we explicitly
+        // use the empty string as spelling.
+        final spelling = '';
+
         parsed.compound.members.add(
           Member(
             dartDoc: getCursorDocComment(
               cursor,
               nesting.length + commentPrefix.length,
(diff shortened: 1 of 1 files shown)


lib/src/header_parser/sub_parsers/compounddecl_parser.dart
```diff
@@ -300,16 +300,25 @@ int _compoundMembersVisitor(clang_types.CXCursor cursor,
         // otherwise they will be added in the next iteration.
         if (!cursor.isAnonymousRecordDecl()) break;
 
+        // From the C11 spec:
```
TzviPM, line 303: Should I spare the details here and instead say:

```
// Anonymous members are always unnamed. To avoid environment-dependent naming issues with the generated code, we explicitly use the empty string as spelling.
```

liamappelbe (approved):

lib/src/header_parser/sub_parsers/compounddecl_parser.dart
liamappelbe: Sure

TzviPM: done

Issue #538, opened 2023-03-21
Dart 3.0 class modifiers


dcharkes:
When code opts into Dart 3.0, we need to add class modifiers.

All subtypes of `Struct`, `Union`, `Opaque`, and `AbiSpecificInteger` need to become `final class` instead of class.

If the code is being generated for pre-3.0, we need to generate `class`, because `final class` is not accepted by the parser.

This means we have two options:

1. Tell people to use an older version of FFIgen if they want to target an older version of the SDK. (We often bump the lower SDK constraint when targeting new `dart:ffi` features anyway.)
2. Add an optional language-version config var, which determines what we output.

I'm not sure of option 2 is worth it though. For example `VarArgs` only exists in 3.0 and we will add that as well when Dart 3.0 is out.

dcharkes:
I'll go for option 1.

We could bump FFIgen to 8.0.0 so that we can still release legacy versions on 7.x. However, that would mean that users need to opt in by explicitly upgrading FFIgen.
Using 7.3.0 instead would mean it auto-upgrades for users. For users on old SDKs the pub version solver would keep solving to 7.2.x because of the 3.0 sdk constraint in 7.3.0.

@leafpetersen  is bumping the SDK constraint to 3.0 and adding class modifiers considered a breaking change? Should it we be major version bumping FFIgen?

leafpetersen:
Bumping the SDK constraint should be fine - the new version will only get picked up by people on 3.0 sdks.  Note though that code generators have some additional complexity:
  - If you are generating code that will be interpreted as being part of a different package, you need to ensure that the generated code explicitly sets its language version, since the package you are generating into still may be on 2.X (and hence by default your generated code would be interpreted in version 2.X).
  -  If you are generating code into other libraries as part files, you need to match the language version of the library into which you are generating

As for whether adding `final` to the generated classes is breaking, it likely is technically breaking in that someone could be extending/implementing the generated classes.  If you feel that someone doing so is clearly violating the intent of your API (and especially if that intent is documented somewhere) you could make the argument that this should not be treated as a breaking change.  As always, I would focus on the actual practical impact on users here.

cc @natebosch @devoncarew @jakemac53 who might have additional advice and opinions about how we want to manage the roll out of class modifiers in dart owned packages and code generators.

oprypin:
[...]

PR #553, opened 2023-04-18
3.0 Class modifiers

dcharkes:
Closes:

* https://github.com/dart-lang/ffigen/issues/538


liamappelbe:
What happens at the moment if a user tries to subclass a subtype of Struct? Does it break, or is it just a weird thing to do that we want to discourage? If it actually breaks, then I'm fine with this change, but if it's just a weird thing to do then Lasse's advice is to avoid unnecessarily restricting what the user does. I'll approve the PR anyway so you're not blocked for a whole timezone cycle.

liamappelbe (approved):

mannprerak2 (approved):

Issue #579, opened 2023-05-16
[ObjectiveC] Invalid bindings generated for AVFAudio framework


TzviPM:
# Context

I was playing around with the objective_c example and using a glob for the entrypoint: `AVFAudio.framework/**.h` instead of `AVFAudio.framework/Headers/AVAudioPlayer.h`. I noticed a few issues with the generated bindings.

# Issues

There are 2 issues that I noticed with the bindings (#578 and #577), and 1 more general issue with some of the interfaces being included:

## Objective C interfaces are included, even if not referenced from an included interface with `exclude-all-by-default: true`

I'm not sure why the `AVAudioNode` is included, except perhaps that there's a section in `AVAudioPlayer.h`:
```c++
#if __has_include(<AVFAudio/AVAudioSession.h>)
#import <AVFAudio/AVAudioSession.h>
#endif
```
And, since I am using a glob pattern for the entry points, `AVAudioSession.h` was included and **it** indirectly references `AVAudioNode`. I think the way we're filtering out declarations may need to be more granular. This also could reduce the size of generated bindings, in general.

Issue #577, opened 2023-05-16
[ObjectiveC] exceptionalReturn of 0 for a function that returns `Float`.


TzviPM:
I was playing around with the objective_c example and using a glob for the entrypoint: `AVFAudio.framework/**.h` instead of `AVFAudio.framework/Headers/AVAudioPlayer.h`. I noticed a few issues with the generated bindings.

There are 2 cases of generated bindings that include some sort of dart function trampoline of type `ffi.Float Function(ffi.Pointer<_ObjCBlock> block, ffi.Pointer<ObjCObject> arg0)` and pass in `0` as a value for `exceptionalReturn`. The resulting error looks like this:

> **Warning**
> The type 'int' must be a subtype of 'Float' for 'fromFunction'.
> Try changing one or both of the type arguments. dart(must_be_a_subtype)

PR #580, opened 2023-05-16
Bug repro for #577, #578, #579

TzviPM:
I was playing around with the objective_c example and using a glob for the entrypoint: `AVFAudio.framework/**.h` instead of `AVFAudio.framework/Headers/AVAudioPlayer.h`. I noticed a few issues with the generated bindings: See #577, #578, and #579


dcharkes:
cc @liamappelbe (mainly responsible for ObjectiveC support).

@TzviPM feel free to contribute fixes for the relevant bugs.

PR #586, opened 2023-06-03
Add ffigen schema and refer them in example config yaml using modeline

### LINKED ISSUE #569: Create a schema for auto-completion of `ffigen.yaml`
https://json-schema.org/

mannprerak2:
Closes #569 

- Added a schema file
- Refer to schema locally using yaml modeline in example configs.


mannprerak2:
I've put the schema at the root level in `ffigen.schema.json`. I am not sure how we proceed though.

mannprerak2:
Some concerns-
1. We could add `additionalProperties: false` but I feel it may be misleading since I'm not sure about the delay in publishing new schema and when the user updates their package.
2. Different ffigen versions could have different schema's, is there any way we can handle this?

ffigen.schema.json
```diff
@@ -0,0 +1,404 @@
+{
+    "$id": "https://json.schemastore.org/ffigen",
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "description": "Configuration for generating dart ffi bindings via ffigen",
+    "type": "object",
```
dcharkes, line 5: Can this file be generated from our configuration spec in Dart?

The benefits:

1. We can hit generate when we add config logic.
2. Either all properties are working or wrong, making it less likely to have typos in this spec.
3. We can generate the schema on the CI, checking it's up to date.

dcharkes (commented):
Sweet!

Thanks @mannprerak2 !

dcharkes:
> Some concerns-
> 
> 1. We could add `additionalProperties: false` but I feel it may be misleading since I'm not sure about the delay in publishing new schema and when the user updates their package.

That's a good question. I don't really know. What's the behavior? Showing an error or a warning? If it's just a warning, maybe it's fine. Because if you mistype a property you would want to see a warning/error.

> 2. Different ffigen versions could have different schema's, is there any way we can handle this?

They don't change that quickly right?

I have no idea if the the pub people cc @jonasfj @sigurdm and dartcode @dantup can tell us something about versioning JSON schemas in vscode via pub packages.

> 3. How do we test this?

See comment above about generating the schema.

@dantup might know how to test VSCode things on the GitHub CI.

DanTup:
Does VS Code support schemas for YAML? There is a RedHat extension (https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml) that says it supports them, but I don't think it's built-in.

In which case users would need to install that extension (and configure it, unless the schema is on schemastore.org). I'm not sure we can reasonably make any of that automatic though.

It might be possible to write some tests that set up VS Code, install the RedHat YAML extension and configure it to check that diagnostics/completion work, although I think if the scheme is valid then the rest of that is already tested by the RedHat extension.

HosseinYousefi:
Installing RedHat YAML extension is required. The schemas are downloaded from https://www.schemastore.org/json/ by default. So we need to upload them there eventually.

Screenshot 2023-06-05 at 15 32 20

---

As @dcharkes mentioned, I'd suggest having this generated automatically.

dcharkes:
Following the discussions in https://github.com/dart-lang/ffigen/issues/569, I'm happy to land a schema, manually publish it to the schema store, have a stable url of the schema on this repo, and have all the examples point to the one on this repo or the schemastore (maybe pointing to a stable URL on this repo would be best because we can guarantee it to stay up to date).

We should probably version the schema file with the semantic version of the package. https://github.com/SchemaStore/schemastore/blob/master/CONTRIBUTING.md

mannprerak2:
> Following the discussions in https://github.com/dart-lang/ffigen/issues/569, I'm happy to land a schema, manually publish it to the schema store, have a stable url of the schema on this repo, and have all the examples point to the one on this repo or the schemastore (maybe pointing to a stable URL on this repo would be best because we can guarantee it to stay up to date).

Cool

For the auto-generation, I'll try to come up with a good abstraction which can handle validation and json schema generation. Will probably need to rewrite `Map<List<String>, Specification> _getSpecs`

dcharkes:
> For the auto-generation, I'll try to come up with a good abstraction which can handle validation and json schema generation. Will probably need to rewrite `Map<List<String>, Specification> _getSpecs`

👍 

This is probably a general pattern we want, I've filed https://github.com/dart-lang/tools/issues/112 to make a general solution. Feel free to either come up with a general solution, and try it both in package:ffigen _and_ package:native_assets_cli, or to make something package:ffigen specific, and report your experiences on that bug.

mannprerak2:
@dcharkes This is not yet ready for a full review, I might do more refactorings, but feel free to take a quick look at the changes and let me know what you think.

Apart from generating the JSON schema programmatically, this PR rewrites most of the config validation and extraction part.

Some considerations
1. Validation might become tighter with these changes (Some validation errors were silently ignored earlier). Maybe we increase the major version to account for this?
2. Should we generate additionalProperties flag as `false`? This will depend on the delay in updating the schema.
3. How much of the description from our configuration should we add to the Schema?
4. The Schema implementation still depends on Yaml. Perhaps the integration with `package:cli_config` could be done at the `result`/`resultOrDefault` closures? (Should probably be handled in a separate PR though)
5. Log output might have changed slightly, especially where OneOfSchema is used.

dcharkes:
> @dcharkes This is not yet ready for a full review, I might do more refactorings, but feel free to take a quick look at the changes and let me know what you think.

Works like a charm locally when I remove some of the required properties! 💪 🎉 🚀 

> Apart from generating the JSON schema programmatically, this PR rewrites most of the config validation and extraction part.

Nice!

Regarding the new schema abstraction. How does it relate to for example https://pub.dev/packages/json_schema or https://pub.dev/packages/json_schema2 ?

* Do some of the classes do the same thing as we're doing? If yes, can we reuse those classes instead of adding implementation to this package?
* If the above answer is no, can our classes output objects of those classes?

I'm trying to ask the high approach/architecture questions here. If we can't reuse any of it, that would be fine, I just want to make sure.

> Some considerations
> 
> 1. Validation might become tighter with these changes (Some validation errors were silently ignored earlier). Maybe we increase the major version to account for this?

Sounds good. 👌 

> 2. Should we generate additionalProperties flag as `false`? This will depend on the delay in updating the schema.

Yes. If we add new things, and users get an error, they can double check.
But if we don't add this property and users misspell optional properties, they will not be notified.

I think for us to make a PR to publish a new version of the schema to the schema repository is not that hard.

> 3. How much of the description from our configuration should we add to the Schema?

We could consider moving everything from the markdown table in README.md to the schema and generating the table as well.

> 4. The Schema implementation still depends on Yaml. Perhaps the integration with `package:cli_config` could be done at the `result`/`resultOrDefault` closures? (Should probably be handled in a separate PR though)

No concerns about addressing that in a follow up PR.

> 5. Log output might have changed slightly, especially where OneOfSchema is used.

No concerns from my side.

> 6. How do we test the generated JSON schema? probably something we can solve together with pushing JSON schema to the schemastore.

We write dart unit tests with https://pub.dev/packages/json_schema 🙂

lib/src/config_provider/schema.dart
```diff
@@ -0,0 +1,732 @@
...
+  /// Extract SchemaNode from [value]. This will call the [transform] for all
+  /// underlying Schemas if valid.
+  /// Should ideally only be called if [validate] returns True. Throws
+  /// [SchemaExtractionError] if any validation fails.
+  SchemaNode extract(dynamic value) {
+    return extractNode(SchemaNode(path: [], value: value));
+  }
+}
+
+class FixedMapKey<DE extends dynamic> {
```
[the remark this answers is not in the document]
mannprerak2: > Why extends dynamic instead of DE extends Object

We must use `Object?` (to support null). And that is the same as `dynamic` I suppose.


> Is it because the YAML and JSON abstractions also all use dynamic in the API so we don't have to write casts everywhere?

I initially did `FixedMapKey<DE>`, but the analyzer complained about specifying the generic type. I realized I had to just supply `<dynamic>` on most places. In general, the more nested configurations with different type of keys are, more `<dynamic>` is likely used in generic.

E.g, you can only do `FixedMapSchema<T>` if every value schema for each subkey, either returns T by itself or is converted to T using the `transform` closure.

Currently the Generic Type in the `Schema` is only used to enforce the type returned by calling extractNode, and also makes the value type-safe in the `transform` closure.

mannprerak2:
> Regarding the new schema abstraction. How does it relate to for example https://pub.dev/packages/json_schema or https://pub.dev/packages/json_schema2 ?

They are more focused on having a JsonSchema input than generating a JsonSchema, The classes probably can't be reused (and might change since they aren't part of the public API). Validations are possible if we write the Schema by hand and use that as an initial step.

The custom validation, extraction and default spec part would still remain, so I feel the current approach might be better.

mannprerak2:
I think this is ready for review now @dcharkes

lib/src/config_provider/config.dart
```diff
@@ -215,363 +219,689 @@ class Config {
...
+          resultOrDefault: (node) => _language = node.value as Language,
+        ),
+        FixedMapKey(
+            key: strings.headers,
+            required: true,
+            valueSchema: FixedMapSchema<List<String>>(
+              keys: [
+                FixedMapKey(
+                  key: strings.entryPoints,
+                  valueSchema: ListSchema<String>(childSchema: StringSchema()),
```
dcharkes, line 274: `ListSchema<String>(childSchema: StringSchema())` String is mentioned twice here.

Can we infer the type argument of `ListSchema` from the `childSchema`?

lib/src/config_provider/config.dart
```diff
@@ -215,363 +219,689 @@ class Config {
...
+  Schema _getRootSchema() {
+    return FixedMapSchema(
+      keys: [
+        FixedMapKey(
+          key: strings.llvmPath,
+          valueSchema: ListSchema<String>(
+            childSchema: StringSchema(),
+            transform: (node) => llvmPathExtractor(node.value),
+          ),
+          defaultValue: (node) => findDylibAtDefaultLocations(),
```
dcharkes, line 231: This is not part of the schema, but is part of the config.

So in some sense we could have some kind of `ConfigSpecification` abstraction that extends/implements `Schema` or has a method that returns a `Schema`. If we go for a method that returns the `Schema`, we need two class hierarchies that mirror each other, that would be verbose. If we go for `extends` we end up with the issue that all the child nodes only have a `Schema` type and not a `ConfigSpecification` type. If we go for `implement`s then all the `Schema` abstract classes would have abstract getters for their children (but that still means two type hierarchies).

Maybe the best thing is to call the whole thing `ConfigSpecification` instead of `Schema`? A config specification needs to specify the data structure (schema), but can specify more. And then our config schema currently just happens to mirror the different schema options and naming conventions.

The naming would then be something like:

* `Schema` -> `ConfigSpecification`
* `FixedMapSchema` -> `MapConfigSpecification` (with an optional argument `Iterable<K>? validKeys`, if no list is provided, you can have arbitrary keys, if a list is provided it's a fixed map?)
* `ListSchema<V>` -> `ListConfigSpecification<V>` and we can restrict values with an optional `Iterable<V>? validValues` argument to make it consistent with `Config.string` from `package:cli_config`.
* `FixedMapKey` -> `MapConfigSpecificationEntry<K,V>`
* `OneOfSchema` -> `OneOfConfigSpecification`

lib/src/config_provider/config.dart
```diff
@@ -215,363 +219,689 @@ class Config {
...
+            ],
+            transform: (node) => ffiNativeExtractor(node.value),
+          ),
+          defaultValue: (node) => FfiNativeConfig(enabled: false),
+          resultOrDefault: (node) =>
+              _ffiNativeConfig = (node.value) as FfiNativeConfig,
+        ),
+      ],
+    );
+  }
```
dcharkes, line 659: Do we have any if key `a` is in this map, then `b` is required?
Or if `a` is in this map, then `b` must _not_ be supplied?

That's probably not expressible in JSON schemas.

But I have some of that kind of logic in `package:native_assets_cli`. https://github.com/dart-lang/native/blob/main/pkgs/native_assets_cli/lib/src/model/build_config.dart

We would probably model that by referring from one schema to the other. I believe the setup in this PR would work fine with trying to add that.

dcharkes (commented):
So happy to see red squigglies in my editor! 🎉

ffigen.schema.json
```diff
@@ -0,0 +1,523 @@
+{
```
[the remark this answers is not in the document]
mannprerak2: We can use [`$comment`](https://json-schema.org/draft-07/json-schema-core.html#rfc.section.9) instead.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 9.0.0
+
+- Updated validation logic to be more strict.
```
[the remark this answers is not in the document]
mannprerak2: Ahh! forgot the main thing xD

lib/src/config_provider/config.dart
mannprerak2: We can't, since child schema might have a `transform` closure, which could change the return value from `extractNode` to  anything (`dynamic`).

The `Generic` is only used for casting the value returned by each of the child objects. In case of ListSchema, it returns a list `<E>[]`. For FixedMapSchema, it returns a map `<dynamic, E>{}`.

dcharkes: Should it have two type arguments then? One for the thing being read, and one for the value after it's transformed.

It's rather common to have a list of valid values that is some kind of enum class.

```dart
          BuildMode.fromString(
            config.string(
              BuildMode.configKey,
              validValues: BuildMode.values.map((e) => '$e'),
            ),
          );
```

https://github.com/dart-lang/native/blob/main/pkgs/native_assets_cli/lib/src/model/build_config.dart

That would then be a `ConfigSpecification` with `<String, BuildMode>`. E.g. `String` is the type in JSON/YAML and `BuildMode` is the type in Dart once parsed/transformed.

lib/src/config_provider/config.dart
mannprerak2: We can use `OneOfSchema` for these situations. It might be too complicated though for a lot of if-else, Since we'd need one Schema for each permutation.

mannprerak2: Another option supported by JsonSchema as well would be https://json-schema.org/understanding-json-schema/reference/conditionals.html#id6

lib/src/config_provider/config.dart
mannprerak2: How about `ConfigSpec` instead? a tad little less to type.

lib/src/config_provider/config.dart
mannprerak2: > Should it have two type arguments then? One for the thing being read, and one for the value after it's transformed.

Sure

> It's rather common to have a list of valid values that is some kind of enum class.

We can simply nest EnumSchema inside ListSchema -
```dart
ListSchema(childSchema: EnumSchema(allowedValues: ["a", "b", "c"]))
```

lib/src/config_provider/config.dart
dcharkes: Ah nice! If I find some time, I can give your abstraction a spin and add these, and use it in package:native_assets_cli. If the abstraction works well standalone, we could move it to package:config_cli. 👌

lib/src/config_provider/config.dart
mannprerak2: I've used `ConfigSpec` for now. 

> MapConfigSpecification (with an optional argument Iterable<K>? validKeys, if no list is provided, you can have arbitrary keys, if a list is provided it's a fixed map?)

This works, but we lose the ability to validate different key patterns with different types of schema (albeit we don't yet have a usecase of different patterns with different schemas, but we might in the future)

dcharkes: image

`ConfigDefinition` would be shorter. Or `ConfigStructure`. Or, .. we call this thing the `Config` and we call the instantiated one the `ConfigInstance`.

@lrhn to the rescue? 😄

mannprerak2: >ListConfigSpecification<V> and we can restrict values with an optional Iterable<V>? validValues argument to make it consistent with Config.string from package:cli_config

We can use `ListSchema(childSchema: EnumSchema(allowedValues: [...]))`

lib/src/config_provider/config.dart
mannprerak2: > Should it have two type arguments then? One for the thing being read, and one for the value after it's transformed.

Done. The first generic casts the value received before calling transform, and the second parameter casts the value after transform or if transform is not available, casts the same result itself.
The second generic basically decides the value returned by _extractNode.

So if no transform is specified, both the Generic types must be equal

pubspec.yaml
```diff
@@ -24,10 +24,11 @@ dependencies:
...
   glob: ^2.0.0
-  file: ^6.0.0
+  file: ^7.0.0
   package_config: ^2.1.0
   yaml_edit: ^2.0.3
 
 dev_dependencies:
   lints: ^2.0.1
   test: ^1.16.2
+  json_schema: ^5.1.1
```
HosseinYousefi, line 34: `json_schema` has a LICENSE that I personally am not familiar with: https://pub.dev/packages/json_schema/license

I don't know what are the implications of using it but I think we should investigate. cc/ @dcharkes

lib/src/config_provider/config.dart
```diff
@@ -215,363 +219,699 @@ class Config {
...
-      } else if (spec!.requirement == Requirement.yes) {
-        _logger.severe("Key '$key' is required.");
-        result = false;
-      } else if (spec.requirement == Requirement.prefer) {
-        _logger.warning("Prefer adding Key '$key' to your config.");
-      }
+  ConfigSpec _getRootSchema() {
+    return FixedMapConfigSpec(
+      entries: [
+        FixedMapEntry(
```
dcharkes, line 225: Remove of `Fixed`. `MapEntryConfigSpec`? (The rest also is ConfigSpec.)

lib/src/config_provider/config.dart
```diff
@@ -215,363 +219,699 @@ class Config {
...
-      if (checkKeyInYaml(key, map)) {
-        result = result && spec!.validator(key, getKeyValueFromYaml(key, map));
-      } else if (spec!.requirement == Requirement.yes) {
-        _logger.severe("Key '$key' is required.");
-        result = false;
-      } else if (spec.requirement == Requirement.prefer) {
-        _logger.warning("Prefer adding Key '$key' to your config.");
-      }
+  ConfigSpec _getRootSchema() {
+    return FixedMapConfigSpec(
```
dcharkes, line 223: Remove `Fixed`, whether it allows extra keys or not should be a bool in the constructor.

lib/src/config_provider/config.dart
```diff
@@ -215,363 +219,699 @@ class Config {
...
+      entries: [
+        FixedMapEntry(
+          key: strings.llvmPath,
+          valueConfigSpec: ListSchema<String, String>(
```
dcharkes, line 227: ListSchema -> ListConfigSpec

lib/src/config_provider/config.dart
```diff
@@ -215,363 +219,699 @@ class Config {
...
+                macIncludeStdLib: ((node.value)[strings.macos]
+                    as Map?)?[strings.includeCStdLib] as bool,
+              ),
+              result: (node) => _compilerOpts.addAll(
+                  (node.value as CompilerOptsAuto).extractCompilerOpts()),
+            )),
+        FixedMapEntry(
+          key: strings.libraryImports,
+          valueConfigSpec:
+              DynamicMapConfigSpec<String, Map<String, LibraryImport>>(
```
dcharkes, line 328: Can we merge `DynamicMapConfigSpec` with `FixedMapConfigSpec`? And just have a boolean field and that field in the constructor?

It feels that conceptually these really should have the same API and implementation except for whether the keys are defined.

On difference is that all keys and values in the dynamic one need to have the same type while this is not true for the fixed one. If that is a more fundamental difference, then the two classes should be called `Homogeneous` and `Hetrogeneous` probably. However, that can also be expressed with the type argument for the values being `dynamic` for the hetrogeneous one.

Are there any other reasons why these two types are not merged?

lib/src/config_provider/config.dart
```diff
@@ -215,363 +219,699 @@ class Config {
...
+                  defaultValue: (node) => Includer.excludeByDefault(),
+                ),
+                FixedMapEntry(
+                  key: strings.leafFunctions,
+                  valueConfigSpec: _includeExcludeObject(),
+                  defaultValue: (node) => Includer.excludeByDefault(),
+                ),
+                FixedMapEntry(
+                  key: strings.varArgFunctions,
+                  valueConfigSpec: _functionVarArgsSchema(),
```
dcharkes, line 363: Schema -> ConfigSpec ? Please do a ctrl+f on Schema to find all the old naming.

Since it's a private method it doesn't matter that much, but still cleaner to clean it up.

lib/src/config_provider/config_spec.dart
```diff
@@ -0,0 +1,811 @@
...
+/// are completed.
+///
+/// Extraction -
+///
+/// - The data is first validated, if invalid it throws a
+/// ConfigSpecExtractionError.
+/// - The extracted data from the child(s) is collected and the value is
+/// transformed via [transform] (if specified).
+/// - Finally the [result] closure is called (if specified).
+abstract class ConfigSpec<TE extends Object?, RE extends Object?> {
```
dcharkes, line 102: This is the most important class, move it to be first in the file?

lib/src/config_provider/config_spec.dart
```diff
@@ -0,0 +1,811 @@
+import 'package:logging/logging.dart';
+import 'package:yaml/yaml.dart';
+
+final _logger = Logger('ffigen.config_provider.config');
+
+/// A container object for a ConfigSpec Object.
+class ConfigSpecNode<TE> {
```
dcharkes, line 7: Should this be:

```dart
/// An individual value in a config for a specific [path] instantiated from a [ConfigSpec].
///
/// A config value contains both the [value] that users of the configuration would want
/// to use, as well as the [rawValue] that was provided as input to the configuration.
class ConfigValue<TE> {
```

(e.g. what is it a container for? I guess the rawValue. However, it also contains the path.)

lib/src/config_provider/config_spec.dart
```diff
@@ -0,0 +1,811 @@
...
+  String get pathString => path.join(" -> ");
+
+  /// Contains the underlying node value after all transformations and
+  /// default values have been applied.
+  final TE value;
+
+  /// Contains the raw underlying node value. Would be null for fields populated
+  /// but default values
+  final Object? rawValue;
+
```
dcharkes, line 25: Would it be useful to have `ConfigSpec` as a field here?

Maybe not, as we use it as an immutable data class with `copyWith`. All the logic is in other classes. Instead of a `ConfigValue` knowing how to transform itself because it has access to the config spec.

lib/src/config_provider/config_spec.dart
```diff
@@ -0,0 +1,811 @@
...
+}
+
+/// ConfigSpec for a Map that can have any number of keys.
+///
+/// [CE] typecasts result from keyValueConfigSpecs->{}->valueConfigSpec.
+///
+/// [RE] typecasts result returned by this node.
+class DynamicMapConfigSpec<CE extends Object?, RE extends Object?>
+    extends ConfigSpec<Map<dynamic, CE>, RE> {
+  /// [keyRegexp] will convert it's input to a String before matching.
```
dcharkes, line 382: nit: This doesn't discribe the field, but how the keys of the map will be used.

dcharkes (approved):

lib/src/config_provider/config_spec.dart
mannprerak2: Well, I used the Tree -> TreeNode analogy, but I guess this seems better.

lib/src/config_provider/config_spec.dart
mannprerak2: Yeah, I intended this to act as an immutable data class. Maybe equivalent to `context` in some ways?

lib/src/config_provider/config_spec.dart
mannprerak2: Updated.

lib/src/config_provider/config.dart
mannprerak2: One of the reasons they are separate is handling default values via FixedMapEntry.

1. We could merge the two, but then have 2 separate lists? One -`final List<FixedMapEntry> entries;` and another - `final List<({String keyRegexp, ConfigSpec valueConfigSpec})> keyValueConfigSpecs;`
2. Or we could keep one class, but have 2 different types of MapEntry - `FixedMapEntry` and `DynamicMapEntry`. So that we can separate these. Any class with a DynamicMapEntry in list would also automatically have `additionalProperties` set to true.

I like 2 better. 

> then the two classes should be called Homogeneous and Hetrogeneous probably. However, that can also be expressed with the type argument for the values being dynamic for the hetrogeneous one.

We could just rename them, but the type argument being dynamic would remove any validations

dcharkes: Okay, maybe we should just keep them separate, even though both of them are Maps.

https://www.google.com/search?q=heterogeneous+map
https://www.google.com/search?q=heterogeneous+map+java

It looks like `HeterogeneousMap` is the right name.

Maybe `HomogenousMap` should just be `Map`, since all maps `<K, V>` are homogeneous already.

(If it makes sense and there's duplicated code, it could be in a private base class of both.)

mannprerak2: Um, I don't have much opinion over the names, so would the new naming be -
FixedMapConfigSpec -> HeterogeneousMapConfigSpec
DynamicMapConfigSpec -> HomogenousMapConfigSpec
FixedMapEntry -> HeterogeneousMapEntry

Also technically even DynamicMap can have different values (so not exactly homogenous), so basically a key with a particular pattern must have a value that conforms to a specified ConfigSpec

dcharkes: `FixedMapConfigSpec` -> `HeterogeneousMapConfigSpec`
`DynamicMapConfigSpec` -> `MapConfigSpec` (because the way the `<K,V>` works here is just how normal maps work)
`FixedMapEntry` -> `HeterogeneousMapEntry`

dcharkes:
Sorry for the 1 million iterations here! 🙈  I think we're ready to land after the renames. 🚀

mannprerak2:
> Sorry for the 1 million iterations here! 🙈 I think we're ready to land after the renames. 🚀

Ah, no problem. It's finally ready now I suppose 🙈

dcharkes:
Thanks @mannprerak2 !

PR #593, opened 2023-07-22
Fix doc comment missing on struct/union array fields, update test, ve…

### LINKED ISSUE #592: Comments are not extracted into generated binding for C array
When generating bindings using ffigen for a C `struct` that contain an `array`, the associated comment of the `array` is not extracted to the generated bindings.

For example if ffigen is run on the following C `struct`:
```c
/// Bug demo structure
struct MyStruct {
  /// Comment for an int is correctly extracted
  int myInt;
  
  /// Comment for an array is not extracted
  int myArray[42];
}
```

... with following pubspec.yaml config:
```yaml
ffigen:
  # ...
  comments:
    style: any
    length: full
```

... the generated binding is:
```dart
/// Bug demo structure
final class MyStruct extends ffi.Struct {
  /// Comment for an int is correctly extracted
  @ffi.Int()
  external int myInt;

  @ffi.Array.multi([42])
  external ffi.Array<ffi.Int> myArray;
}
```

I would have expected the comment `/// Comment for an array is not extracted` to be extracted to the generated binding dart file.
This is the case with both ffigen 8.0.2 and 9.0.0 (I have not performed tests on other versions) with flutter 3.10.6.

mannprerak2:
Closes #592

### DIFF
--- a/lib/src/header_parser/sub_parsers/functiondecl_parser.dart
+++ b/lib/src/header_parser/sub_parsers/functiondecl_parser.dart
@@ -70,8 +70,8 @@ List<Func>? parseFunctionDeclaration(clang_types.CXCursor cursor) {
 
     // Initialized with a single value with no prefix and empty var args.
     var varArgFunctions = [VarArgFunction('', [])];
-    if (clang.clang_isFunctionTypeVariadic(cursor.type()) == 1) {
-      if (config.varArgFunctions.containsKey(funcName)) {
+    if (config.varArgFunctions.containsKey(funcName)) {
+      if (clang.clang_isFunctionTypeVariadic(cursor.type()) == 1) {
         varArgFunctions = config.varArgFunctions[funcName]!;
       } else {
         _logger.warning(
--- a/test/header_parser_tests/comment_markup.h
+++ b/test/header_parser_tests/comment_markup.h
@@ -23,4 +23,7 @@ struct Com4{
 
     /* Single line field comment. */
     float b;
+
+    /* Comment on array member */
+    int c[3];
 };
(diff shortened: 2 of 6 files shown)


lib/src/header_parser/sub_parsers/functiondecl_parser.dart
```diff
@@ -70,8 +70,8 @@ List<Func>? parseFunctionDeclaration(clang_types.CXCursor cursor) {
 
     // Initialized with a single value with no prefix and empty var args.
     var varArgFunctions = [VarArgFunction('', [])];
-    if (clang.clang_isFunctionTypeVariadic(cursor.type()) == 1) {
-      if (config.varArgFunctions.containsKey(funcName)) {
+    if (config.varArgFunctions.containsKey(funcName)) {
+      if (clang.clang_isFunctionTypeVariadic(cursor.type()) == 1) {
```
mannprerak2, line 74: Just discovered this by accident when parsing -
```C
extern void abcd();
```
```txt
[WARNING]: Skipping variadic-argument config for function abcd since its not variadic.
```

This didn't affect the flow, but generated a misleading log for every variadic function there would be in the source code

test/header_parser_tests/comment_markup.h
```diff
@@ -23,4 +23,7 @@ struct Com4{
 
     /* Single line field comment. */
     float b;
+
+    /* Comment on array member */
```
dcharkes, line 27: nit: period at end of sentence.

dcharkes (approved):
Thanks @mannprerak2 !

test/header_parser_tests/comment_markup.h
mannprerak2: Fixed

PR #594, opened 2023-07-24
Allow extern inline functions to be generated.

### LINKED ISSUE #588: [C] "Inline" function available in compiled dylib, but no bindings generated by ffigen?
I'm currently trying to bind a bitset implementation found in CRoaring to Dart via ffigen.

As expected, ffigen doesn't seem to generate bindings for inline functions such as:

```
static inline size_t bitset_size_in_words(const bitset_t *bitset) {
    return bitset->arraysize;
}
```

As expected, this function is not visible in the compiled dylib when searched for via `nm -gU libroaring.dylib`

However, if the static is changed to `extern`, i.e.:

```
extern inline size_t bitset_size_in_words(const bitset_t *bitset) {
    return bitset->arraysize;
}
```

Then the function becomes available in the compiled dylib:

```
> nm -gU libroaring.dylib | grep bitset_size_in_words         
> 0000000000001b08 T _bitset_size_in_words
```

but ffigen still won't generate bindings for it despite it being available in the dylib.

I'm wondering what's the exact limitation here? If a function is available in the dylib, shouldn't ffigen be able to generate appropriate bindings for it (regardless of whether it is inline annotated or not?)

Cf. https://github.com/RoaringBitmap/CRoaring/issues/487

mannprerak2:
Closes #588

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 9.0.2
+
+- Allow extern inline functions to be generated.
+
 # 9.0.1
 
 - Fix doc comment missing on struct/union array fields.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,7 @@
-# 9.0.2
-
-- Allow extern inline functions to be generated.
-
 # 9.0.1
 
 - Fix doc comment missing on struct/union array fields.
+- Allow extern inline functions to be generated.
 
 # 9.0.0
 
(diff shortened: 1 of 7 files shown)


mannprerak2:
@dcharkes should I keep the version as 9.0.1 and merge the changelog?

dcharkes:
> @dcharkes should I keep the version as 9.0.1 and merge the changelog?

Yes we can, I didn't hit the publish button yet.

mannprerak2:
Done

PR #599, opened 2023-08-25
Bump CI SDK version to 3.1.0

liamappelbe:


dcharkes:
This just bumps the CI, not the minimum SDK constraint, is that intended?

liamappelbe:
> is that intended?

Yeah, I don't actually want to land this, I'm just testing stuff. I'm seeing lots of errors on my machine, and they looked like they were related to some newer Dart features. So I wanted to see if the same errors show up on github CI if I bump the version. But nope, CI is working, so there's something wrong on my end.

PR #601, opened 2023-08-29
Listener/Async Blocks

liamappelbe:
Add support for listener blocks, built on `NativeCallable.listener`. Also, bump the SDK version to 3.2 so we get the `keepIsolateAlive` field (otherwise the internal async trampoline will prevent the isolate from closing).

All the substantive changes are in lib/src/code_generator/objc_block.dart. `Block.listener` uses all the same infrastructure as `Block.fromFunction`, even using the exact same `registerClosure` and `closureTrampoline` functions. The only difference is that listeners wrap the `closureTrampoline` in a `NativeCallable.listener`, whereas ordinary Blocks wrap that same function in a `Pointer.fromFunction`.

### DIFF
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        sdk: [3.1.0]
+        sdk: [3.2.0-42.1.beta]
     steps:
       - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9
       - uses: dart-lang/setup-dart@d6a63dab3335f427404425de0fbfed4686d93c4f
@@ -44,7 +44,7 @@ jobs:
       - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9
       - uses: dart-lang/setup-dart@d6a63dab3335f427404425de0fbfed4686d93c4f
         with:
-          sdk: 3.1.0
+          sdk: 3.2.0-42.1.beta
       - name: Install dependencies
         run: dart pub get
       - name: Install libclang-14-dev
@@ -61,7 +61,7 @@ jobs:
       - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9
       - uses: dart-lang/setup-dart@d6a63dab3335f427404425de0fbfed4686d93c4f
         with:
-          sdk: 3.1.0
+          sdk: 3.2.0-42.1.beta
       - name: Install dependencies
         run: dart pub get
       - name: Build test dylib and bindings
@@ -83,7 +83,7 @@ jobs:
       - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9
       - uses: dart-lang/setup-dart@d6a63dab3335f427404425de0fbfed4686d93c4f
         with:
-          sdk: 3.1.0
+          sdk: 3.2.0-42.1.beta
       - name: Install dependencies
         run: dart pub get
       - name: Build test dylib and bindings
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        sdk: [3.2.0-42.1.beta]
+        sdk: [3.2.0-114.0.dev]
     steps:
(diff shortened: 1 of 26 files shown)


test/native_objc_test/block_test.dart
```diff
@@ -64,15 +65,44 @@ void main() {
...
+      int value = 0;
+      final block = ObjCBlock.listener(lib, () {
+        value = 123;
+        hasRun.complete();
+      });
+
+      final thread = BlockTester.callOnNewThread_(lib, block);
+      thread.start();
+
+      await hasRun.future;
```
dcharkes, line 100: Nice! 🚀

lib/src/config_provider/config_spec.dart
```diff
@@ -591,7 +591,7 @@ class StringConfigSpec<RE extends Object?> extends ConfigSpec<String, RE> {
     if (!o.checkType<String>(log: log)) {
       return false;
     }
-    if (_regexp != null && !_regexp!.hasMatch(o.value as String)) {
+    if (_regexp != null && !_regexp.hasMatch(o.value as String)) {
```
dcharkes, line 594: field promotion?

lib/src/code_generator/objc_block.dart
```diff
@@ -115,13 +121,37 @@ class $name extends _ObjCBlockBase {
...
+
+    // Listener block constructor is only available for void blocks.
+    if (isVoid) {
+      s.write('''
+  /// Creates a listener block from a Dart function.
+  ///
+  /// This is based on FFI's NativeCallable.listener, and has the same
+  /// capabilities and limitations. This block can be invoked from any thread,
+  /// but only supports void functions, and is not run synchronously. See
+  /// NativeCallable.listener for more details.
```
dcharkes, line 145: This should probably have some documentation about not keeping the isolate alive.

Also, are there use cases in which users would want to keep the isolate alive? Should we surface a bool on the API?

.github/workflows/test-package.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        sdk: [3.0.0]
+        sdk: [3.2.0-114.0.dev]
```
dcharkes, line 22: Please add a TODO to bump this to 3.2 stable once it is released.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 10.0.0-dev.0
+
+- Add support for ObjC Blocks that can be invoked from any thread, using
+  NativeCallable.listener.
+- Bump min SDK version to 3.2.
```
dcharkes, line 5: `3.2.0-114.0.dev`

We can modify it to 3.2 once we release 10.0.0. (If we ever want to release a dev version for users it would be nice if the changelog would be precise.)

pubspec.yaml
```diff
@@ -3,7 +3,7 @@
 # BSD-style license that can be found in the LICENSE file.
 
 name: ffigen
-version: 9.0.1
+version: 10.0.0-dev.0
 description: Generator for FFI bindings, using LibClang to parse C header files.
```
dcharkes, line 7: This should be updated to "Generator for FFI bindings, using LibClang to parse C, Objective-C and Swift files."

It's probably fair enough to say that we support Objective-C and Swift now.

dcharkes (approved):
LGTM.

Should we maybe add some integration test with a real world API that needs this?

(Nit, I stand by https://github.com/dart-lang/ffigen/pull/363#issuecomment-1124560479, I now had to review the generator instead of the generated API. 👓 )

lib/src/config_provider/config_spec.dart
liamappelbe: Better fix: https://github.com/dart-lang/ffigen/pull/605

pubspec.yaml
```diff
@@ -13,7 +15,7 @@ topics:
  - codegen
 
 environment:
-  sdk: ">=3.0.0 <4.0.0"
+  sdk: ">=3.2.0-114.0.dev <4.0.0"
```
vaind, line 18: What's the reason for the generator (ffigen) to require the new SDK version?

liamappelbe: Listener blocks are built on `NativeCallable.listener`. While that feature did land in 3.1, there are some important updates to it in 3.2 that we need (specifically `keepIsolateAlive = false`).

PR #605, opened 2023-08-31
Fix null inference weirdness

liamappelbe:
The change to lib/src/config_provider/config_spec.dart in #601 only works on the very latest SDK snapshot. This will make it hard for people to contribute until it hits stable. Better to just rewrite the line to sidestep the inference.

### DIFF
--- a/lib/src/config_provider/config_spec.dart
+++ b/lib/src/config_provider/config_spec.dart
@@ -591,7 +591,7 @@ class StringConfigSpec<RE extends Object?> extends ConfigSpec<String, RE> {
     if (!o.checkType<String>(log: log)) {
       return false;
     }
-    if (_regexp != null && !_regexp.hasMatch(o.value as String)) {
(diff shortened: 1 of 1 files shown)


lib/src/config_provider/config_spec.dart
```diff
@@ -591,7 +591,7 @@ class StringConfigSpec<RE extends Object?> extends ConfigSpec<String, RE> {
     if (!o.checkType<String>(log: log)) {
       return false;
     }
-    if (_regexp != null && !_regexp.hasMatch(o.value as String)) {
+    if (!(_regexp?.hasMatch(o.value as String) ?? true)) {
```
dcharkes, line 594: nit: does the `!` bind stronger than the `??`, I guess it does. Maybe add parenthesis.

dcharkes (approved):
lgtm

Thanks @liamappelbe !

PR #609, opened 2023-09-01
Use correct exceptional return in float blocks

### LINKED ISSUE #577: [ObjectiveC] exceptionalReturn of 0 for a function that returns `Float`.

liamappelbe:
We were generating `0` as the exceptional return for float blocks, but we should be generating `0.0`. Fixes #577

### DIFF
--- a/lib/src/code_generator/imports.dart
+++ b/lib/src/code_generator/imports.dart
@@ -86,8 +86,8 @@ final unsignedLongLongType =
     ImportedType(ffiImport, 'UnsignedLongLong', 'int', '0');
 final longLongType = ImportedType(ffiImport, 'LongLong', 'int', '0');
 
-final floatType = ImportedType(ffiImport, 'Float', 'double', '0');
-final doubleType = ImportedType(ffiImport, 'Double', 'double', '0');
+final floatType = ImportedType(ffiImport, 'Float', 'double', '0.0');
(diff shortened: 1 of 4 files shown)


dcharkes (approved):
lgtm!

Thanks @liamappelbe .

Should we have a bug-fix entry in the CHANGELOG.md?

Also, should we have fixes on the stable branch? 9.0.2 ?

liamappelbe:
> Should we have a bug-fix entry in the CHANGELOG.md?
> 
> Also, should we have fixes on the stable branch? 9.0.2 ?

I have a lot of small bug fixes like this in the works. Maybe we can cherry pick them into a stable branch once they're done, if 3.2 isn't stable by then.

PR #607, opened 2023-08-31
fix: objc block trampoline invalid return type

vaind:
Fixes #591

As for updating the generated `avf_audio_bindings.dart` - I have done so locally but haven't committed because I've got a lot of unrelated changes - probably due to the different environment to the one this was originally generated on.

---

- [x] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>
  <summary>Contribution guidelines:</summary><br>

- See our [contributor guide](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md) for general expectations for PRs.
- Larger or significant changes should be discussed in an issue before creating a PR.
- Contributions to our repos should follow the [Dart style guide](https://dart.dev/guides/language/effective-dart) and use `dart format`.
- Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#making-a-change).
- Changes to packages require [corresponding tests](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md#Testing).

Note that many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback.
</details>

PR #613, opened 2023-09-04
Use stret and fpret variants of objc_msgSend where appropriate

### LINKED ISSUE #608: Use `objc_msgSend_stret` and `objc_msgSend_fpret` instead of `objc_msgSend` for signature that require that
Repro: https://github.com/dnys1/ffigen_crash

We received a [crash report](https://github.com/aws-amplify/amplify-flutter/issues/3630) which I traced back to this code:

```dart
final screen = NSScreen.getMainScreen(lib);
screen!.visibleFrame;
//           ^ crashes on access to visibleFrame
```

The issue is not reproducible on Silicon macOS and was reported from a user with an Intel machine. I was able to reproduce it by running the repro app above under Rosetta, which I learned can be enabled in Xcode as a run destination under **Product** > **Destination** > **Destination Architectures** > **Show Both**.

Both myself and the user are running on Flutter 3.13 and Xcode `14.3.1`. This issue is reproducible on Flutter 3.10 and Flutter beta as well.

Thanks for any help!

liamappelbe:
Fixes #608

Also add changelog entries for #609 and #607.


lib/src/code_generator/objc_built_in_functions.dart
```diff
@@ -396,3 +390,37 @@ class ObjCInternalGlobal extends LookUpBinding {
     binding?.addDependencies(dependencies);
   }
 }
+
+/// A wrapper around the objc_msgSend function, or the stret or fpret variants.
+class ObjCMsgSendFunc {
+  final bool isStret;
```
dcharkes, line 396: Nit: It seems kind of weird we have 3 variants but then a bool.

Can we have an enum, and then use it for setting `originalName` as well?

dcharkes (approved):
lgtm!

Thanks @liamappelbe !

PR #618, opened 2023-09-08
Fix ObjC `instancetype` inheritance issue.

### LINKED ISSUE #486: Some objects are being classified as NSObject
I generated bindings for classes in the Apple HealthKit framework, but some of the objects are being classified as NSObject. 

For example:

pubspec.yaml:
```
ffigen:
  name: HealthKit
  description: "Create bindings for the HealthKit APIs"
  language: objc
  output: "lib/healthkit_bindings.dart"
  compiler-opts:
    - '-F/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/System/Library/Frameworks'
  exclude-all-by-default: true
  objc-interfaces:
    include:
      - "HKHealthStore"
      - "HKQuery"
  headers:
    entry-points:
      - "/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/System/Library/Frameworks/HealthKit.framework/Headers/HKHealthStore.h"
      - "/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/System/Library/Frameworks/HealthKit.framework/Headers/HKQuery.h"
 ```

```
 var healthStore = hk.HKHealthStore.alloc(lib).init();
```

`healthStore` is an `NSObject` as opposed to a `HKHealthStore`. 

This can be fixed with:

```
var healthStore = hk.HKHealthStore.castFrom(hk.HKHealthStore.alloc(lib).init());
```

This is for a tutorial for Apple developers, so should the general guidance be that `.castFrom()` is used in places where dart is recognizing the object as `NSObject`? Is it considered a bug or expected behavior?

### PR DESCRIPTION
**Pull request by liamappelbe**
`instancetype` is special in ObjC. It's only allowed as the return type of a method, and always takes on the type of the enclosing class, even when that method is being inherited from a base class. So in the generated Dart code we need to copy the method to the subclass, so that we can `@override` it and return the subclass, rather than just inheriting the base class's version, which returns the base class.

Fixes #486

### DIFF
--- a/test/native_objc_test/inherited_instancetype_test.dart
+++ b/test/native_objc_test/inherited_instancetype_test.dart
@@ -1,0 +1,57 @@
+// Copyright (c) 2023, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// Objective C support is only available on mac.
+@TestOn('mac-os')
+
+// Regression tests for https://github.com/dart-lang/ffigen/issues/486
+
+import 'dart:ffi';
+import 'dart:io';
+
+import 'package:test/test.dart';
+import '../test_utils.dart';
+import 'inherited_instancetype_bindings.dart';
+import 'util.dart';
+
+void main() {
+  late InheritedInstancetypeTestObjCLibrary lib;
+
+  group('inheritedInstancetype', () {
+    setUpAll(() {
+      logWarnings();
+      final dylib = File('test/native_objc_test/inherited_instancetype_test.dylib');
+      verifySetupFile(dylib);
+      lib = InheritedInstancetypeTestObjCLibrary(DynamicLibrary.open(dylib.absolute.path));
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: test/native_objc_test/inherited_instancetype_test.dart [hunk 1]
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) 2023, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// Objective C support is only available on mac.
+@TestOn('mac-os')
+
+// Regression tests for https://github.com/dart-lang/ffigen/issues/486
```
dcharkes, line 8: Nit: period at the end of sentence.

dcharkes (approved):
lgtm

Thanks @liamappelbe ! 🚀

PR #621, opened 2023-09-15
Better typedefs for `NativeFunction`s

### LINKED ISSUE #614: Ffigen doesn't generate typedefs compatible with NativeCallable's APIs
Hello!

Today I was trying out the new [NativeCallable](https://api.dart.dev/stable/3.1.0/dart-ffi/NativeCallable-class.html) APIs and stumbled upon on a kinda annoying issue.

NativeCallable T generic parameter has this problem with ffigen that it requires specifying a function that only uses dart:ffi types, but there isn't currently a way to get one of these, since the generated typedefs usually come from pointer functions.

I think the NativeCallable API should not need to be changed, but say, a static function that removes the pointer and nativefunction from the signature would be useful? Although I do believe that the API should have bound T type to extend [NativeFunction](https://api.dart.dev/stable/3.1.0/dart-ffi/NativeFunction-class.html). 

I haven't checked if ffigen's config section functions -> expose-typedefs would generate the required typedefs, but if it does, it should be documented in the NativeCallable class too.

liamappelbe:
When generating typedefs for `Pointer<NativeFunction<Function>>`, also  generate a typedef for the `Function`.

Fixes #614.

### DIFF
--- a/example/libclang-example/generated_bindings.dart
+++ b/example/libclang-example/generated_bindings.dart
@@ -9751,6 +9751,8 @@ typedef CXCursorVisitor = ffi.Pointer<
     ffi.NativeFunction<
         ffi.Int32 Function(
             CXCursor cursor, CXCursor parent, CXClientData client_data)>>;
+typedef CXCursorVisitor_function = ffi.Int32 Function(
+    CXCursor cursor, CXCursor parent, CXClientData client_data);
 
 /// Describes how the traversal of the children of a particular
 /// cursor should proceed after visiting a particular child cursor.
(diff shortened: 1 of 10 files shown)


example/libclang-example/generated_bindings.dart
```diff
@@ -10432,6 +10434,11 @@ typedef CXInclusionVisitor = ffi.Pointer<
             ffi.Pointer<CXSourceLocation> inclusion_stack,
```
dcharkes, line 10434: These typedefs should refer to the new ones instead of duplicating them.

example/libclang-example/generated_bindings.dart
```diff
@@ -10432,6 +10434,11 @@ typedef CXInclusionVisitor = ffi.Pointer<
...
             ffi.UnsignedInt include_len,
             CXClientData client_data)>>;
+typedef CXInclusionVisitor_function = ffi.Void Function(
```
dcharkes, line 10437: Naming-wise, it feels like this one should be without `_function` and the other one should be renamed `_pointer`. Do you know if we have a lot of uses of these typedefs, if it would be worth the breaking change?

dcharkes (approved):
lgtm

example/libclang-example/generated_bindings.dart
liamappelbe: Not sure, but we could avoid a breaking change by putting this behind a flag. Eg, add a flag called `use-function-typedefs`, and if it's true we add `_pointer` to the existing definition and then define the function definition with the current name. WDYT?

dcharkes: Let's just merge as is.

Issue #531, opened 2023-03-08
Objective-C blocks should use the generated Dart class rather than ffi.Pointer<ObjCObject>


brianquinlan:
Currently a method like:

```C
- (void)readDataOfMinLength:(NSUInteger)minBytes 
                  maxLength:(NSUInteger)maxBytes 
                    timeout:(NSTimeInterval)timeout 
          completionHandler:(void (^)(NSData *data, BOOL atEOF, NSError *error))completionHandler;
```

Will generate a block like:
```dart
class ObjCBlock40 extends _ObjCBlockBase {
...
  void call(
      ffi.Pointer<ObjCObject> arg0, bool arg1, ffi.Pointer<ObjCObject> arg2) {
...
```

Instead, `call` should have a signature like:
```dart
class ObjCBlock40 extends _ObjCBlockBase {
...
  void call(
      NSData data, bool atEOF, NSError error) {
...
```

Using the correct argument names is for bonus points ;-)

Issue #500, opened 2022-12-21
C-style functions consuming Objective-C classes should allow the Objective-C class to be passed


brianquinlan:
A function declaration like:
```C
void foo(NSDate *date);
```

Generates this code:

```dart
void foo(ffi.Pointer<ObjCObject> date) {
  ...
}
```

Instead, it could generate code like:

```dart
void foo(NSDate date) {
  ...
}
```

Or at least there should be an option or variant that accepts the `NSDate`.

brianquinlan:
@leighajarett

brianquinlan:
One other thing to note is that `foo` from above is not callable given a `NSDate` because the `ffi.Pointer<ObjCObject>` in the wrapped `NSDate` is private

i.e.

```dart
class _ObjCWrapper implements ffi.Finalizable {
  final ffi.Pointer<ObjCObject> _id;
  ...
}

Issue #386, opened 2022-05-25
ObjC: Unnecessarily abstract types generated for some typedefs


brianquinlan:
For example, this Objective-C code:

```
typedef NSString *NSErrorDomain;

...
@interface NSError : NSObject <NSCopying, NSSecureCoding>
...
@property (readonly, copy) NSErrorDomain domain;
...
@end
```

Generates this code:
```dart
typedef dispatch_queue_t1 = ffi.Pointer<ObjCObject>;
typedef NSErrorDomain = ffi.Pointer<ObjCObject>;
typedef NSErrorUserInfoKey = ffi.Pointer<ObjCObject>;

class NSError extends NSObject {
  ...
  /// These define the error. Domains are described by names that are arbitrary strings used to differentiate groups of codes; for custom domain using reverse-DNS naming will help avoid conflicts. Codes are domain-specific.
  NSErrorDomain get domain {
    return _lib._objc_msgSend_1(_id, _lib._sel_domain1);
  }
  ...
```

A more natural typedef  would be:
```dart
typedef NSErrorDomain = NSString
```

liamappelbe:
Looking into this, and it's a bit of a rabbit hole. I'm going to have to do a series of related PRs that fixes this, #309, #531, and #500. All these problems boil down to the difference between the Dart type passed through FFI (`Pointer<ObjCObject>`) and the wrapper object (`NSString`).

As an example of why it's more complicated than just changing the typedef:

```C
typedef NSString *NSErrorDomain;

struct SomeStruct {
  NSErrorDomain error_domain;
};
```

Would generate this Dart:

```Dart
// This is the typedef we want.
typedef NSErrorDomain = NSString;

class SomeStruct extends Struct {
  // But that typedef is invalid here, because we're trying to use
  // the Dart wrapper object as a struct field.
  external NSErrorDomain error_domain;
}
```

Instead we have to generate getters and setters:

```Dart
typedef NSErrorDomain = NSString;

class SomeStruct extends Struct {
  // Need a private field with the correct type.
  external Pointer<ObjCObject> _error_domain;

  // And getters and setters that handle conversion. Due to the
  // lib arg, the getter can't be a real getter. Probably should
  // make the setter not real too for symmetry.
  NSErrorDomain get_error_domain(NativeLibrary lib) =>
      NSErrorDomain._(_error_domain, lib, retain: true, release: true);
  void set_error_domain(NSErrorDomain value) => _error_domain = value._id;
}
```

[...]

PR #623, opened 2023-09-21
Separate getDartType and getUserType

liamappelbe:
A bunch of ObjC polish bugs are related to the fact that we don't draw a distinction between the 2 different uses of `Type.getDartType`:

1. The Dart type used in FFI function signatures and structs. The type that `Type.getCType` maps to.
2. The user facing type we present in the generated API.

For C bindings these 2 types are always the same. But for ObjC types they can be different. For example, an `NSString*` has a Dart type of `Pointer<ObjCObject>` and a user type of the Dart wrapper class `NSString`. I've been distinguishing them in objc_interface.dart with some hard coded special cases for class methods, but this misses top level functions, ObjC blocks, struct fields, and typedefs.

More info: https://github.com/dart-lang/ffigen/issues/386#issuecomment-1728584216

So we need to separate these 2 uses by defining a new `Type.getUserType` method. Over time I'll migrate the rest of ffigen to make this distinction (ie find everywhere `Type.getDartType` is called and switch the ones that are user facing to `Type.getUserType`), and that will fix all these bugs. But it'll be complicated because we have to do conversions between the Dart type and the user type. So to keep things simple I've just migrated objc_interface.dart for now.

To facilitate this I also had to define new specializations of some `Type`s:

- `ObjCObject*` was an ordinary `Pointer` but is now `ObjCObjectPointer`
- `instancetype` was a `Typealias` but is now `ObjCInstanceType`

Related bugs: #309 #386 #531 #500

### DIFF
--- a/lib/src/code_generator/objc_interface.dart
+++ b/lib/src/code_generator/objc_interface.dart
@@ -5,7 +5,6 @@
 import 'package:ffigen/src/code_generator.dart';
 import 'package:logging/logging.dart';
 
-import '../strings.dart' as strings;
 import 'binding_string.dart';
 import 'utils.dart';
 import 'writer.dart';
--- a/lib/src/code_generator/objc_interface.dart
+++ b/lib/src/code_generator/objc_interface.dart
@@ -281,7 +281,7 @@ class $name extends ${superType?.name ?? '_ObjCWrapper'} {
       if (m.isClass &&
           !_excludedNSObjectClassMethods.contains(m.originalName)) {
         addMethod(m);
-      } else if (_isInstanceType(m.returnType)) {
+      } else if (m.returnType is ObjCInstanceType) {
         addMethod(m);
       }
     }
@@ -355,13 +355,8 @@ class $name extends ${superType?.name ?? '_ObjCWrapper'} {
   @override
   String getCType(Writer w) => PointerType(objCObjectType).getCType(w);
 
-  bool _isObject(Type type) =>
-      type is PointerType && type.child == objCObjectType;
-
-  bool _isInstanceType(Type type) =>
-      type is Typealias &&
-      type.originalName == strings.objcInstanceType &&
-      _isObject(type.type);
+  @override
+  String getUserType(Writer w) => name;
 
   // Utils for converting between the internal types passed to native code, and
   // the external types visible to the user. For example, ObjCInterfaces are
@@ -370,16 +365,12 @@ class $name extends ${superType?.name ?? '_ObjCWrapper'} {
   bool _needsConverting(Type type) =>
       type is ObjCInterface ||
       type is ObjCBlock ||
-      _isObject(type) ||
-      _isInstanceType(type);
+      type is ObjCObjectPointer ||
+      type is ObjCInstanceType;
 
   String _getConvertedType(Type type, Writer w, String enclosingClass) {
-    if (type is BooleanType) return 'bool';
-    if (type is ObjCInterface) return type.name;
-    if (type is ObjCBlock) return type.name;
-    if (_isObject(type)) return 'NSObject';
-    if (_isInstanceType(type)) return enclosingClass;
-    return type.getDartType(w);
+    if (type is ObjCInstanceType) return enclosingClass;
+    return type.getUserType(w);
   }
 
   String _getConvertedReturnType(
@@ -393,8 +384,8 @@ class $name extends ${superType?.name ?? '_ObjCWrapper'} {
 
   String _doArgConversion(ObjCMethodParam arg) {
     if (arg.type is ObjCInterface ||
-        _isObject(arg.type) ||
-        _isInstanceType(arg.type) ||
+        arg.type is ObjCObjectPointer ||
+        arg.type is ObjCInstanceType ||
         arg.type is ObjCBlock) {
       if (arg.isNullable) {
         return '${arg.name}?._id ?? ffi.nullptr';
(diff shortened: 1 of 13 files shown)


lib/src/code_generator/type.dart
```diff
@@ -39,10 +39,15 @@ abstract class Type {
   /// passed to native code.
   String getCType(Writer w) => throw 'No mapping for type: $this';
 
-  /// Returns the Dart type of the Type. This is the user visible type that is
-  /// passed to Dart code.
+  /// Returns the Dart type of the Type. This is the type that is passed from
+  /// FFI to Dart code.
   String getDartType(Writer w) => getCType(w);
```
dcharkes, line 44: > * `getCType`
> * `getDartType`
> * `getUserType`

I think I'd prefer the naming as:

* `getCType`
* `getFfiDartType` / `getTrampolineDartType` / `getDartUnwrappedType` / `getInternalDartType`
* `getDartType`

Keeping `getDartType` to be the user visible type.

lib/src/code_generator/pointer.dart
```diff
@@ -9,7 +9,15 @@ import 'writer.dart';
 /// Represents a pointer.
 class PointerType extends Type {
   final Type child;
-  PointerType(this.child);
+
+  PointerType._(this.child);
+
+  factory PointerType(Type child) {
```
dcharkes, line 15: Should we maybe have something more general to recognize "internal-dart-types" and return the "user-visible-dart-types"?

For example, if we have something more general we could also recognize `Pointer<Char>` and make it available as `String` in signatures?

It should probably be a separate pass. Rather than making a choice point in factory constructors of types.

```dart
  // Parse the bindings according to config object provided.
  final library = parse(config);

  final libraryTransformed = transform(library);

  // Generate file for the parsed bindings.
  final gen = File(config.output);
  libraryTransformed.generateFile(gen);
```

Or maybe because we already make choices in `parse` based on the config, it should a `Type wrapDartType(Type type)` function being invoked on any type. And for now it only replaces Pointer<CObject> with ObjCObjectPointer.

lib/src/code_generator/typealias.dart
```diff
@@ -122,3 +135,30 @@ class Typealias extends BindingType {
   String? getDefaultValue(Writer w, String nativeLib) =>
       type.getDefaultValue(w, nativeLib);
 }
+
+/// Objective C's instancetype.
+///
+/// This is an alias for an NSObject* that is special cased in code generation.
```
dcharkes, line 141: Maybe explain the special casing.

Add a reference to the Objective-C documentation for `instancetype`.

lib/src/code_generator/objc_interface.dart
```diff
@@ -370,16 +364,12 @@ class $name extends ${superType?.name ?? '_ObjCWrapper'} {
...
+      type is ObjCInstanceType;
 
   String _getConvertedType(Type type, Writer w, String enclosingClass) {
-    if (type is BooleanType) return 'bool';
-    if (type is ObjCInterface) return type.name;
-    if (type is ObjCBlock) return type.name;
-    if (_isObject(type)) return 'NSObject';
-    if (_isInstanceType(type)) return enclosingClass;
-    return type.getDartType(w);
+    if (type is ObjCInstanceType) return enclosingClass;
```
dcharkes, line 371: Nice cleanup!

This special case is the `instancetype` right? The one that returns the type of the object itself even in subclasses.

lib/src/code_generator/typealias.dart
```diff
@@ -122,3 +135,30 @@ class Typealias extends BindingType {
...
+class ObjCInstanceType extends Typealias {
```
dcharkes, line 142: Would it be possible to populate the enclosing class here? So that the code generator doesn't make the decision but we do it here?

dcharkes (changes requested):
> https://github.com/dart-lang/ffigen/issues/386#issuecomment-1728584216

Nice write up!

Reading your write-up, I think we should make a general solution for where the user-visible-dart-type is not equal to the internally-used-with-ffi-dart-type.

Should we have a separate pass `parse` -> `transform` -> `generate` where in the transform step we recognize unwrapped types and wrap them? The current PR special cases the Pointer constructor, but that seems not the right place for the decision point.

cc @mannprerak2 What do you think about this? 

> That migration should be incremental to keep the PR size reasonable.

I'd prefer not having the repo in a half-migrated state. I'm happy with incremental PRs, but then please commit to doing the follow up PRs soon. (I want to avoid having a half-migrated repo and then contributors (including ourselves) having to deal with that technical debt.) Having a full migration PR is preferable imo.

lib/src/code_generator/type.dart
liamappelbe: Done

liamappelbe:
> I'd prefer not having the repo in a half-migrated state.

Well, the migration isn't really incremental. This PR fully migrates all existing bits of the repo that made this distinction (only objc_interface.dart). The incremental part is applying that distinction to the rest of the code gen. Each time I change one of those other modules to distinguish `getFfiDartType` and `getDartType`, I'll be fixing one of these ObjC polish bugs. So I think it makes sense to do them as follow up PRs. But I plan to do them immediately after this one.

dcharkes (approved):
lgtm!

I think we should consider refactoring this so that it becomes more of a pipeline with individual steps being responsible for individual aspects. Now it starts to be a spaghetti with lots of responsibilities both in the parse and in the generate step. However, we should not gate this CL on that.

dcharkes:
Thanks @liamappelbe ! 🚀

PR #629, opened 2023-10-09
Refactor `sameFfiDartAndCType` to not require a `Writer`

liamappelbe:
Ran into this while working on other refactors, and pulled it out into its own PR. Instead of `sameFfiDartAndCType` comparing `getFfiDartType` and `getCType`, just make it a first class method on `Type` that subclasses have to override. This means we don't need a `Writer`, so we can use this function outside of `toBindingString`. If I did it right, there should be no codegen differences.


dcharkes (approved):
lgtm! 🚀

PR #632, opened 2023-10-25
Use getDartType rather than getFfiDartType in ObjC block codegen

### LINKED ISSUE #531: Objective-C blocks should use the generated Dart class rather than ffi.Pointer<ObjCObject>

liamappelbe:
Improves the codegen for blocks with arguments or return types that are objects, nullables, or blocks. Users can now pass or receive the Dart wrapper object, rather than dealing with pointers.

Fixes #531

### DIFF
--- a/lib/src/code_generator/type.dart
+++ b/lib/src/code_generator/type.dart
@@ -54,8 +54,24 @@ abstract class Type {
   /// Returns whether the dart type and C type string are same.
   bool get sameDartAndCType => sameFfiDartAndCType;
 
-  /// Returns the string representation of the Type, for debugging purposes
-  /// only. This string should not be printed as generated code.
+  /// Returns generated Dart code that converts the given value from its
+  /// DartType to its FfiDartType.
+  String convertDartTypeToFfiDartType(Writer w, String value) => value;
+
+  /// Returns generated Dart code that converts the given value from its
+  /// FfiDartType to its DartType.
+  String convertFfiDartTypeToDartType(
+    Writer w,
+    String value,
+    String library, {
+    bool isObjCOwnedReturn = false,
+    String? objCEnclosingClass,
+  }) =>
+      value;
+
+  /// Returns a human readable string representation of the Type. This is mostly
+  /// just for debugging, but it may also be used for non-functional code (eg to
+  /// name a variable or type in generated code).
   @override
   String toString();
 
@@ -106,6 +122,19 @@ abstract class BindingType extends NoLookUpBinding implements Type {
 
   @override
   bool get sameDartAndCType => sameFfiDartAndCType;
+
+  @override
+  String convertDartTypeToFfiDartType(Writer w, String value) => value;
+
+  @override
+  String convertFfiDartTypeToDartType(
+    Writer w,
+    String value,
+    String library, {
+    bool isObjCOwnedReturn = false,
+    String? objCEnclosingClass,
+  }) =>
+      value;
 
   @override
   String toString() => originalName;
--- a/lib/src/code_generator/typealias.dart
+++ b/lib/src/code_generator/typealias.dart
@@ -151,6 +151,26 @@ class Typealias extends BindingType {
   bool get sameDartAndCType => type.sameDartAndCType;
 
   @override
+  String convertDartTypeToFfiDartType(Writer w, String value) =>
+      type.convertDartTypeToFfiDartType(w, value);
+
+  @override
+  String convertFfiDartTypeToDartType(
+    Writer w,
+    String value,
+    String library, {
+    bool isObjCOwnedReturn = false,
+    String? objCEnclosingClass,
+  }) =>
+      type.convertFfiDartTypeToDartType(
+        w,
+        value,
+        library,
+        isObjCOwnedReturn: isObjCOwnedReturn,
+        objCEnclosingClass: objCEnclosingClass,
+      );
+
+  @override
   String cacheKey() => type.cacheKey();
 
   @override
@@ -173,4 +193,18 @@ class ObjCInstanceType extends Typealias {
     super.genFfiDartType,
     super.isInternal,
   }) : super._();
+
+  @override
+  String convertDartTypeToFfiDartType(Writer w, String value) => '$value._id';
+
+  @override
+  String convertFfiDartTypeToDartType(
+    Writer w,
+    String value,
+    String library, {
+    bool isObjCOwnedReturn = false,
+    String? objCEnclosingClass,
+  }) =>
+      ObjCInterface.generateConstructor(
+          objCEnclosingClass ?? 'NSObject', value, library, isObjCOwnedReturn);
 }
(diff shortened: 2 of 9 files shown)


liamappelbe:
Marking not ready for review. I can't repro the mac CI failure locally. Need to investigate more.

liamappelbe:
The failures were to do with ref counting mismatches. They're all fixed now.

lib/src/code_generator/type.dart
```diff
@@ -54,8 +54,29 @@ abstract class Type {
   /// Returns whether the dart type and C type string are same.
   bool get sameDartAndCType => sameFfiDartAndCType;
 
-  /// Returns the string representation of the Type, for debugging purposes
-  /// only. This string should not be printed as generated code.
+  /// Returns generated Dart code that converts the given value from its
+  /// DartType to its FfiDartType.
+  String convertDartTypeToFfiDartType(
```
dcharkes, line 59: Please document the params so that if someone not familiar with objC introduces a new implementation of `Type` doesn't have to reverse engineer.

lib/src/code_generator/type.dart
```diff
@@ -54,8 +54,29 @@ abstract class Type {
...
+    Writer w,
+    String value, {
+    bool objCShouldRetain = false,
+  }) =>
+      value;
+
+  /// Returns generated Dart code that converts the given value from its
+  /// FfiDartType to its DartType.
+  String convertFfiDartTypeToDartType(
```
dcharkes, line 68: Ditto, please document the arguments.

lib/src/code_generator/type.dart
```diff
@@ -54,8 +54,29 @@ abstract class Type {
...
   bool get sameDartAndCType => sameFfiDartAndCType;
 
-  /// Returns the string representation of the Type, for debugging purposes
-  /// only. This string should not be printed as generated code.
+  /// Returns generated Dart code that converts the given value from its
+  /// DartType to its FfiDartType.
+  String convertDartTypeToFfiDartType(
+    Writer w,
+    String value, {
+    bool objCShouldRetain = false,
```
dcharkes, line 62: `objCRetain` (It's not a should, it's guaranteed, we generate a `retain: true`.)

Also, let's make it non-optional. I think it would be useful that all call sites explicitly reason about memory management in the generated code.

lib/src/code_generator/type.dart
```diff
@@ -54,8 +54,29 @@ abstract class Type {
...
+  }) =>
+      value;
+
+  /// Returns generated Dart code that converts the given value from its
+  /// FfiDartType to its DartType.
+  String convertFfiDartTypeToDartType(
+    Writer w,
+    String value,
+    String library, {
+    bool objCShouldRetain = true,
```
dcharkes, line 72: Ditto, `objCRetain` and required arg.

lib/src/code_generator/typealias.dart
```diff
@@ -173,4 +206,23 @@ class ObjCInstanceType extends Typealias {
...
+  @override
+  String convertFfiDartTypeToDartType(
+    Writer w,
+    String value,
+    String library, {
+    bool objCShouldRetain = true,
+    String? objCEnclosingClass,
+  }) =>
+      ObjCInterface.generateConstructor(
+          objCEnclosingClass ?? 'NSObject', value, library, objCShouldRetain);
```
dcharkes, line 227: Is `NSObject` ever happening?
If so, is it a sign we are not generating a class because it's not part of the `ffigen.yaml`?

If it can happen:
- Please add a test case that exercises this.
  - Leave a comment here to that test case.
- Emit a warning.

If it cannot (or should not) happen:
- Add an assert or exception.

dcharkes (approved):
lgtm w. comments

Thanks @liamappelbe!

(If you want to release a version, don't forget changelog entry and pubspec version bump.)

test/native_objc_test/block_test.m
```diff
@@ -12,11 +12,47 @@
...
+@end
+
+
 typedef int32_t (^IntBlock)(int32_t);
 typedef float (^FloatBlock)(float);
 typedef double (^DoubleBlock)(double);
 typedef Vec4 (^Vec4Block)(Vec4);
 typedef void (^VoidBlock)();
+typedef DummyObject* (^ObjectBlock)(DummyObject*);
+typedef DummyObject* _Nullable (^NullableBlock)(DummyObject* _Nullable);
```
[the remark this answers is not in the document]
liamappelbe: How about NullableObjectBlock?

test/native_objc_test/block_test.m
```diff
@@ -12,11 +12,47 @@
   double w;
 } Vec4;
```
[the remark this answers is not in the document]
liamappelbe: Done

test/native_objc_test/block_test.dart
```diff
@@ -181,10 +251,206 @@ void main() {
...
+    }
+
+    test('Calling a native block block from ObjC has correct ref counting', () {
+      final (blockBlock, outputBlock) = nativeBlockBlockObjCCallRefCountTest();
+      doGC();
+      expect(BlockTester.getBlockRetainCount_(lib, blockBlock), 0);
+      expect(BlockTester.getBlockRetainCount_(lib, outputBlock), 0);
+    });
+
+    (Pointer<Int32>, Pointer<Int32>) objectBlockRefCountTest() {
```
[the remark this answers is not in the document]
liamappelbe: Done

test/native_objc_test/block_test.dart
```diff
@@ -162,6 +165,73 @@ void main() {
...
+
+    test('Nullable block', () {
+      bool isCalled = false;
+      final block = NullableBlock.fromFunction(lib, (DummyObject? x) {
+        isCalled = true;
+        return x;
+      });
+
+      final obj = DummyObject.new1(lib);
+      final result1 = block(obj);
```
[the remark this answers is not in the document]
liamappelbe: Done

lib/src/code_generator/typealias.dart
liamappelbe: Shouldn't ever happen (clang should give an error). Added a null assertion instead.

lib/src/code_generator/type.dart
liamappelbe: Done

lib/src/code_generator/type.dart
liamappelbe: Done.

lib/src/code_generator/type.dart
liamappelbe: Done

lib/src/code_generator/type.dart
liamappelbe: Done.

PR #633, opened 2023-10-31
ObjC static functions

### LINKED ISSUE #500: C-style functions consuming Objective-C classes should allow the Objective-C class to be passed

liamappelbe:
Use Dart wrapper objects when passing args or receiving results from ObjC static functions.

For ref counting, we assume functions do not return a retained reference by default (so we have to retain it ourselves). We also parse the function annotations, and if we see a `NS_RETURNS_RETAINED`, then the function is returning a retained reference, so we don't have to manually retain it in that case.

Fixes #500, #309


test/native_objc_test/static_func_test.m
```diff
@@ -0,0 +1,64 @@
...
+}
+
+- (void)dealloc {
+  if (counter != nil) --*counter;
+  [super dealloc];
+}
+
+@end
+
+StaticFuncTestObj* staticFuncOfObject(StaticFuncTestObj* a) {
```
dcharkes, line 42: Can we move these (the actual functions used in the test) up, and the helper object down? Or does ObjC require an ordering?

test/native_objc_test/static_func_test.dart
```diff
@@ -0,0 +1,156 @@
...
+      generateBindingsForCoverage('static_func');
+    });
+
+    doGC() {
+      final gcNow = "gc-now".toNativeUtf8();
+      executeInternalCommand(gcNow.cast(), nullptr);
+      calloc.free(gcNow);
+    }
+
+    Pointer<Int32> staticFuncOfObjectRefCountTest(Allocator alloc) {
```
dcharkes, line 47: It seems this only has one call site, is this because of scoping that it's in a separate function?

dcharkes (approved):
Lgtm!

Thanks @liamappelbe!

test/native_objc_test/static_func_test.m
liamappelbe: Done. Just need to have the interface decl above them I think.

test/native_objc_test/static_func_test.dart
liamappelbe: Yep. I use the function scoping and `doGC()` to ensure that all the wrapper objects are collected.

PR #634, opened 2023-11-01
Fix generated asset/assetId property for ffi-native

nmfisher:
Using `asset` under `ffi-native` generates an incorrect `asset` property rather than `assetId` ([reference](https://github.com/dart-lang/sdk/blob/0544ecba9a301030a91fae8b399bdae5deb67c0e/sdk/lib/ffi/ffi.dart#L1262C25-L1262C25)), which fails to compile. This PR:
-  changes the generated annotation property from `asset` to `assetId` 
- changes the YAML key from `asset` to `assetId` (technically a breaking change to existing configs, but these were presumably broken anyway)
- adds tests for `ffi-native` 

---

- [ Y ] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>
  <summary>Contribution guidelines:</summary><br>

- See our [contributor guide](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md) for general expectations for PRs.
- Larger or significant changes should be discussed in an issue before creating a PR.
- Contributions to our repos should follow the [Dart style guide](https://dart.dev/guides/language/effective-dart) and use `dart format`.
- Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#making-a-change).
- Changes to packages require [corresponding tests](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md#Testing).

Note that many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback.
</details>


test/ffi_native_test/build_test_dylib.dart
```diff
@@ -0,0 +1,129 @@
+// Copyright (c) 2020, the Dart project authors. Please see the AUTHORS file
```
dcharkes, line 1: If we want to run the native code, can we extend the `native_test` instead of copy pasting code? Adding a second `config.yaml` there.

Also, running the code will not work on all OSes. Only Windows and MacOS will do `DynamicLibrary.open` with global symbol visibility, the test will fail on Linux. https://github.com/dart-lang/sdk/issues/50105#issuecomment-1441764283 `@Native`s are supposed to be used with the native assets feature https://github.com/dart-lang/sdk/issues/50565 which provides the mapping from `@Native`s to the dylib path. Maybe land the PR with only expect code gen tests instead of running the test. We already have a codegen test in test/example_tests/ffinative_example_test.dart.

dcharkes (changes requested):
Thanks @nmfisher! ❤️ 

> technically a breaking change to existing configs, but these were presumably broken anyway

We're bumping the version to 10.0 anyway, so that should be fine.

It might have been that people were using the deprecated `@FfiNative`s, but I doubt it due to the native assets feature not having landed yet.

test/ffi_native_test/build_test_dylib.dart
nmfisher: I wasn't exactly sure if you only wanted to update `test/example_tests/ffinative_example_test.dart` or also move the codegen test into `test/native_test/native_test.dart` so I did both. If that's unnecessary duplication I can just delete the latter.

dcharkes: The latter is unnecessary, it's just testing the same thing right?

nmfisher: Correct, I'll remove.

dcharkes (approved):
Thanks @nmfisher! 🎸

